PR #80, opened 2020-08-05
fix modulo case

### LINKED ISSUE #79: Doesn't render "%" character
vue-18n: v9.0.0.-alpha.11
vue: 3.0.0-rc.5

I use dayjs and i should return my string similar to this:

````
messages: {
  en: {
    time: {
      future: 'in %s',
    }
  }
}
````

I call t() in setup() function:

````
import { useI18n } from 'vue-i18n'

setup () {
  const { t, locale } = useI18n()
  console.log( t('time.future') )
}
````

However, it returns "in s" instead of "in %s".

I am not sure if it is a bug or a feature. How can we fix this?

### PR DESCRIPTION
**Pull request by kazupon**
closes #79

### DIFF
--- a/test/message/compiler.test.ts
+++ b/test/message/compiler.test.ts
@@ -46,6 +46,21 @@ describe('edge cases', () => {
     })
     expect(code.toString()).toMatchSnapshot('code')
   })
+
+  test(`hi %s !`, () => {
+    const code = compile(`hi %s !`)
+    expect(code.toString()).toMatchSnapshot('code')
+  })
+
+  test(`%`, () => {
+    const code = compile(`%`)
+    expect(code.toString()).toMatchSnapshot('code')
+  })
+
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: test/message/compiler.test.ts [hunk 1]
```diff
@@ -46,6 +46,21 @@ describe('edge cases', () => {
...
+    const code = compile(`hi %s !`)
+    expect(code.toString()).toMatchSnapshot('code')
+  })
+
+  test(`%`, () => {
+    const code = compile(`%`)
+    expect(code.toString()).toMatchSnapshot('code')
+  })
+
+  test(`no apples %| one apple % |  too much apples  `, () => {
```
lmiller1990, line 60: Should be too *many* :P

kazupon: you're right, my poor english ... 😓

PR #87, opened 2020-08-07
feat: how to get locale messages from a given key

### LINKED ISSUE #75: Curious why object/array returns from $t are removed?
I use this frequently when defining content that is systematic.

For example when i have components that take objects, for instance, I have a FormControl component that takes an object of `validationMessages` keyed to validation rules. I can rely on vue-i18n to store this object exactly as I want it with no need for tricky back and forth:

```js
// i18n
form: {
  firstNameInput:{
   label: 'first name',
   validationMessages: {
     required: 'please enter a first name',
     length: 'please ask your parents for a shorter name',
   }
 }
}

// component
<FormControl validation-messages="$t('form.firstNameInput.validationMessages)"/>

// or even
<FormControl v-bind="$t('form.firstNameInput)"/>
```

Or when I know something will be a list:

```js
// i18n
cancellation: {
  reasons: [`It was too big`, `It was too small`, `It looked a bit too much like a sailboat`, `it had a strange smell`]
}

// component
<ul>
  <li v-for="reason in $t('cancellation.reasons')>{{ reason }}</li>
</ul>
```

This flexibility has been amazing for us, allowing us to rely on vue-i18n to store structured content that we can use programmatically (especially when defining form controls). I don't understand any benefit to ensuring objects don't come back. Is there a new function that can be used instead?

kazupon:
resolve #75 
add the bellow API:

- Composable API style
  - Composer: tm method
- Legacy API style
  - VueI18n: tm method
  - Injection of Vue prototype: $tm

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,6 @@ See the below examples:
 - [Example with using Composable API](https://github.com/intlify/vue-i18n-next/blob/master/examples/composable/components/datetime-format.html)
 - [Example with using Legacy API](https://github.com/intlify/vue-i18n-next/blob/master/examples/legacy/components/datetime-format.html)
 
-
 ## :lollipop: Examples
 
 See the [`examples`](https://github.com/intlify/vue-i18n-next/tree/master/examples) directory.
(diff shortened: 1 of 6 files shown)


thedamon:
Thank you!! 🙂

PR #94, opened 2020-08-23
refactor: simplify scope cases

meteorlxy:
- When `emptyOption` is `true`, `scope` is always `global`
- When `scope` is `global`, we do not need to call `getCurrentInstance`, so it's better to return earlier than `getCurrentInstance()`

---

BTW, the following snippet might not be necessary:

```ts
  if (instance == null) {
    throw createI18nError(I18nErrorCodes.UNEXPECTED_ERROR)
  }
```

Becasue if the currentInstance is not available, the `inject()` api will return `undefined` and an error has already been thrown:

```ts
  const i18n = inject(I18nSymbol) as I18n<
    Messages,
    DateTimeFormats,
    NumberFormats
  >
  if (!i18n) {
    throw createI18nError(I18nErrorCodes.NOT_INSLALLED)
  }
```

### DIFF
--- a/src/i18n.ts
+++ b/src/i18n.ts
@@ -371,15 +371,14 @@ export function useI18n<
 
   const global = i18n.global
 
-  let emptyOption = false
   // prettier-ignore
-  const scope: I18nScope = (emptyOption = isEmptyObject(options)) // eslint-disable-line no-cond-assign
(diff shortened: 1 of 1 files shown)


kazupon (approved):
Thanks! :)

PR #101, opened 2020-08-29
Fix devtools typo

minddust:
This global var doesn't match it's definition (due to missing `I`) which causes the entire vite (esm) production build to fail with:

```
ReferenceError: __INTLFY_PROD_DEVTOOLS__ is not defined
``` 

fallback defined in `mist.ts`:

``` 
  if (typeof __FEATURE_PROD_DEVTOOLS__ !== 'boolean') {
    needWarn = true
    getGlobalThis().__INTLIFY_PROD_DEVTOOLS__ = false
  }
```


kazupon:
OMG! 
Good Catch!
Thanks!

PR #122, opened 2020-09-25
feat: global props and methods injection for composable mode

### LINKED ISSUE #106: No transaltion for nested or child components
Hello, thank you for your work. I'm trying to use the `vue-i18n-next` with `vue 3` but it does not seem to work for nested components (or child component). 

### vue & vue-i18n version
- vue: 3.0.0-beta.1
- vue-i18n: 9.0.0-alpha.16

### Steps to reproduce 
I tried codesandbox.io but there are some issues with vue 3 import, so here is the complete list of files to reproduce the issue.

In the App wrapper view, there is a i18n text which works and an other component `Login` which also use i18n. But i18n only works for the wrapper container and not for the login view.

##### File `App.vue`
```
<template>
    {{t("login.title")}}
    <br/>
    does it works for login ?
    <Login></Login>
</template>

<script>
        import {useI18n} from "vue-i18n";
	import Login from "./components/Accounts/Login";

	export default {
		name: 'App',
		components: {
			Login
		},
		setup() {
			return useI18n()
		}
	}
</script>

<style>

</style>

```
##### File `Login.vue`
```
<template>
<div>
    {{ t("login.title") }}
</div>
</template>

<script>

	export default {
		name: 'Login',
		components: {
		}
	}
</script>

<style lang="scss">
</style>

```
##### File `main.js`
```
import {createApp} from 'vue'
import App from './App.vue'
import router from "./routes/routes";
import i18n from "./i18n";

const app = createApp(App);

app.use(router);
app.use(i18n);

app.mount('#app')

```

##### File `i18n.js`
```
import {createI18n} from 'vue-i18n'

const i18n = createI18n({
	locale: 'en',
	globalInstall: true,
	messages: {
		en: {'login.title': 'test'}
	}
})
export default i18n;

```

### What is Expected?
I expect to translate the key inside the login view also and nested children.
Only the translation inside `App.vue` works, the translate for `Login.vue` does not work.
#### Screenshots
- UI : 
image
- Console : 
image

kazupon:
resolve #106

This PR is a PR to improve the DX of the API vue-i18n-next provides for the composition API.

This PR improves the following:
- no longer needs to call `useI18n` on each component when using global composer
- The following global props & methods are injected by `app.config.globalProperties`
  - `$i18n` prop: exported `locale`, `fallbackLocale` and `availableLocales` props
  - `$t` method
  - `$d` method
  - `$n` method
  - `$tm` method

### DIFF
--- a/examples/composable/scope/global.html
+++ b/examples/composable/scope/global.html
@@ -10,18 +10,18 @@
     <div id="app">
       <h1>Root</h1>
       <form>
-        <label>{{ t('message.language') }}</label>
-        <select v-model="locale">
+        <label>{{ $t('message.language') }}</label>
+        <select v-model="$i18n.locale">
           <option value="en">en</option>
           <option value="ja">ja</option>
         </select>
       </form>
-      <p>{{ t('message.hello') }}</p>
+      <p>{{ $t('message.hello') }}</p>
       <Child />
     </div>
     <script>
       const { createApp } = Vue
-      const { createI18n, useI18n } = VueI18n
+      const { createI18n } = VueI18n
 
       const SlotChild = {
         template: `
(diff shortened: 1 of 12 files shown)


ux-engineer:
Awesome! Just tomorrow I was refactoring components to Vue 3 and was wondering if I could use 't' easier way than importing and invoking `useI18n()`...

PR #126, opened 2020-09-25
feat: vue-devtools plugin

kazupon:


src/composer.ts
```diff
@@ -213,7 +216,7 @@ export interface Composer<
   ): void
   mergeLocaleMessage(
     locale: Locale,
-    message: LocaleMessageDictionary<Message>
+    uessage: LocaleMessageDictionary<Message>
```
syuilo, line 219: typo?

kazupon: fixed with #131

PR #134, opened 2020-10-03
Fix `UNEXPECTED_ERROR` message typo

jungleBadger:
During my tests I found a typo in one of the error messages, as demonstrated below:

Captura de tela 2020-10-03 175450


exoego (approved):
Nice catch!

kazupon:
good catch!
thanks!

PR #163, opened 2020-10-28
Documentation - t()

michaelzangl:
Currently, the .d.ts is stripped of all comments. There is already some documentation, so i included the comments and added some more.

One thing to note:
I removed the t(...args: unknown[]), since users should only see the typed versions of that method.


package.json
```diff
@@ -70,7 +70,8 @@
     "typescript": "^4.0.2",
     "typescript-eslint-language-service": "^4.0.1",
     "vitepress": "^0.7.0",
-    "vue": "^3.0.0"
+    "vue": "^3.0.0",
+    "yarn": "^1.22.10"
```
kazupon, line 74: please don't include to devDependencies. 🙏 
package manager should install globally your enviroment!

kazupon (changes requested):

kazupon:
@michaelzangl 
vue-i18n-next API docs will be generated with `api-extractor`.
so, we need to be commented with the tsdocs style
https://api-extractor.com/pages/tsdoc/doc_comment_syntax/

michaelzangl:
@kazupon I cannot regenerate the documentation. When i run 'docs:api', it puts the documentation in `docs`, not in `docs/api`. What do you use to re-generate those files? Wouldn't it be better to not include them in the GIT and instead add them as build artifacts somewhere?

kazupon:
I've just re-setup docs enviroment.
you can run the commends:

- `yarn docs:setup`: build typescript codes and extract api markdown docs
- `yarn docs:dev`: run vitepress

kazupon:
I'll merge your PRs.
If you want to continue to update the document, please create a new one!
Thanks for your contribution!

kazupon (approved):

PR #165, opened 2020-10-28
Allow d() to take an ISO string as parameter.

michaelzangl:
Currently, only Date objects and number are allowed.

While using number may cause many problems (missing timezone information), a ISO string is a format returned by most APIs. So in my opinion, it makes more sense to allow an (ISO) string to be passed to this method.

### DIFF
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -200,10 +200,10 @@ export interface Composer<
   t(key: Path, named: NamedValue, defaultMsg: string): string
   t(key: Path, named: NamedValue, options: TranslateOptions): string
   t(...args: unknown[]): string // for internal
-  d(value: number | Date): string
-  d(value: number | Date, key: string): string
-  d(value: number | Date, key: string, locale: Locale): string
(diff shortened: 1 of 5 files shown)


src/composer.ts
```diff
@@ -200,10 +200,10 @@ export interface Composer<
   t(key: Path, named: NamedValue, defaultMsg: string): string
   t(key: Path, named: NamedValue, options: TranslateOptions): string
   t(...args: unknown[]): string // for internal
-  d(value: number | Date): string
-  d(value: number | Date, key: string): string
-  d(value: number | Date, key: string, locale: Locale): string
-  d(value: number | Date, options: DateTimeOptions): string
+  d(value: number | Date | string): string
```
kazupon, line 203: When typescript 4.1 is released, we should define ISO type  use template literal types.
https://devblogs.microsoft.com/typescript/announcing-typescript-4-1-beta/#template-literal-types

kazupon (approved):
Thanks!
Great works! 👍 
LGTM!

PR #178, opened 2020-11-03
fix: can't assign to property "processor" on false: not an object

nrzull:
`translateVNode` crashes in `finally` clause because `_context` variable doesn't exist in scope


kazupon (approved):
Thanks!

PR #190, opened 2020-11-17
fix: locale changing for legacy mode

### LINKED ISSUE #186: can't change local
### vue & vue-i18n version
"vue": "^3.0.2",
"vue-i18n": "^9.0.0-beta.6",
"@intlify/vue-i18n-loader": "^2.0.0-beta.1",

### Steps to reproduce
as mentioned here - https://kazupon.github.io/vue-i18n/guide/locale.html

		methods: {
			changeLanguage: function (language) {
				this.$root.$i18n.locale = language;
			},
		},
### What is Expected?
all  messages should change to selected local

doesn't change

kazupon:
resolves #186

### DIFF
--- a/.textlintrc.js
+++ b/.textlintrc.js
@@ -26,7 +26,8 @@ module.exports = {
         'YAML',
         'JSON5',
         'HTTP',
-        'CLI'
+        'CLI',
+        'SFC'
       ]
     },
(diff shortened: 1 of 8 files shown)


kazupon:
@Alymbek
In about #186, I could find resolve way for that issue.
when next version is released, you will try to check it!

PR #242, opened 2020-12-20
Fix not to get loc when passing true to ParserOptions.location

ota-meshi:
This PR fixes an issue where passing `true` to `ParserOptions.location` would cause the parser not to generate the `loc`.


kazupon (approved):
Thanks!

PR #251, opened 2020-12-22
improvement(message-compiler): Improves parse error messages

ota-meshi:
This PR improves parse error messages.


kazupon (approved):
Thanks! LGTM!

PR #258, opened 2020-12-23
docs: update installing for esm bundler

sh0ber:
Adding an example showing how to `import` from a different build.  This is the one asked about most frequently on Stack Overflow, so I think it's nice to have in this section, but maybe it should go somewhere else.


kazupon (approved):
Thanks!

PR #278, opened 2021-01-06
Fix #259 : fix some problems of example prtoject lazy-loading/webpack

### LINKED ISSUE #259: Lazy loading with vue-router fails
- vue: 3.0.4
- vue-i18n: 9.0.0-beta.16
- vue-router: 4.0.1

### Test repo Link
https://github.com/screets/_i18n-test

### Issues

I've followed both "examples" directory and documentation to apply vue-i18n and vue-router. The language has changed successfully when you do through "select" field found on the header.

- It does NOT work when you call website in another url (i.e. mydomain.com/tr/about)
- It gives fallback errors on browser console although it translate it

Screen Shot 2020-12-23 at 12 08 09

### What is Expected?

No browser errors and should work when you directly request page for another language (i.e. mydomain.com/tr/about)

Any help will highly appreciated!

PeterAlfredLee:
This is a fix for #259, including:
1. return 404 when user directly access page or refresh page if current page's path is not '/'
2. jump to default home page when directly access a specific language page
3. jump to default home page in some case when switch the language label

And do also:
1. redirect to default home page when access a specific language page that does not exit

### DIFF
--- a/examples/lazy-loading/webpack/src/App.vue
+++ b/examples/lazy-loading/webpack/src/App.vue
@@ -35,7 +35,7 @@ export default defineComponent({
     // when change the locale, go to locale route
     watch(locale, val => {
       router.push({
-        name: route.name,
+        name: router.currentRoute._rawValue.name,
         params: { locale: val }
       })
     })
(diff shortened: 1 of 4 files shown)


kazupon:
Thanks!

kazupon (approved):

Issue #271, opened 2020-12-30
Add option to support working with flattened json using dot ('.') as separator


RobinB-Gbanga:
Currently, if I look up this key:
`$t('mainMenu.buttonStart')`

The key is only found if I use a nested json:
```
{
  "mainMenu": {
    "buttonStart": "Start!"
  }
}
```

But not if I use a flat json representation:
```
{
  "mainMenu.buttonStart": "Start!"
}
```

For a current project, we are interfacing with a translation source that can only output flat json files. At the same time, however, we'd like to have the benefits of a namespaced/scoped key naming convention. For example, the [tool we use in VSCode](https://marketplace.visualstudio.com/items?itemName=antfu.i18n-ally) offers a nice tree view when using dots as namespace/scope separators.

It would be great if vue-i18n would offer a setting to support flattened json files (or, in other words, to treat the dot ('.') as a normal/literal character).

Thanks!

kazupon:
Thank you for your feedback.

In order for vue-i18n-next to support this, it needs to support the custom message resolver I/F.

Since the message resolver is modularized, we need to extend the custom message resolver so that it can be optioned by vue-i18n.
https://github.com/intlify/vue-i18n-next/tree/master/packages/message-resolver

kazupon:
Fixed with #294

kazupon:
Re-open, so #294 is breaking change.
https://github.com/intlify/vue-i18n-next/issues/271

kazupon:
Sorry late my reply

This has been resolved through multiple features that were added since this issue was opened:

### 1. `flatJson` option

Automatically converts flat keys into nested structures at setup time:

```ts
createI18n({
  flatJson: true,
  messages: {
    en: {
      'mainMenu.buttonStart': 'Start!'
    }
  }
})

t('mainMenu.buttonStart')  // → "Start!"
```

### 2. Custom `messageResolver`

You can provide a custom resolver that treats dots as literal characters:

```ts
createI18n({
  messageResolver: (obj, path) => obj[path] ?? null,
  messages: { ... }
})
```

### 3. Flat key fallback in `t()` and `te()`

In v12 (still alpha), both `t()` and `te()` now fall back to direct object key lookup (`message[key]`) when the nested path resolver returns null. This means flat keys like `"mainMenu.buttonStart"` work without any special configuration.

Closing this as resolved.
Thanks!

PR #294, opened 2021-01-16
Add support for flat json separated with dot('.')

PeterAlfredLee:
This is an implemention for #271.

With `handleFlatJson`, we can recrusively search the object for keys containing dot('.'), and transform the flat structure to normal structure so we can handle them as before.

The corresponding tests are also included.

### DIFF
--- a/packages/message-resolver/src/index.ts
+++ b/packages/message-resolver/src/index.ts
@@ -321,3 +321,50 @@ export function resolveValue(obj: unknown, path: Path): PathValue {
 
   return last
 }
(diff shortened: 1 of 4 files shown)


kazupon (approved):
Thank you for your PR!
This is great works!
LGTM!

PeterAlfredLee:
I'm glad this helps a little bit. :-)

tangjinzhou:
This is a break change. we should support configurable. And not enabled by default.

PR #306, opened 2021-01-20
fix `hasOwnProperty is not a function`

### LINKED ISSUE #305: Uncaught TypeError: obj.hasOwnProperty is not a function
https://github.com/intlify/vue-i18n-next/blob/9e6f6ea3f068e1c467f1781fc264b1d9d37ba996/packages/message-resolver/src/index.ts#L336

This breaks for my app after updating to rc3, rc2 was fine.

Im using 
```
createI18n({
  messages: MESSAGES,
  locale: DEFAULT_LOCALE,
  fallbackLocale: DEFAULT_LOCALE,
})
```

with 
```
export const MESSAGES = {
  [Locales.EN]: enAPP,
  [Locales.DE]: deAPP,
}
```

The locale files contain normal json like this
```
{
  "tests": {
    "main": "Hello there general kenobi"
  }
}
```

Hopefully you can figure that out, any help appreciated.

Edit: reference screenshot for better understanding
image

PeterAlfredLee:
Fix #305

We should use function ` Object.prototype.hasOwnProperty.call` instead of function `hasOwnProperty` to check property.
Sorry for not consider enough before.

I notice there is a function [hasOwn](https://github.com/intlify/vue-i18n-next/blob/dcd005c238151b0924163b679992185755785af7/packages/vue-i18n/src/composer.ts#L973) in `packages/vue-i18n/src/composer.ts`.
This PR would like to move function `hasOwn` to `/packages/shared/src/index.ts` and replace all `hasOwnProperty` call with `hasOwn` call.

### DIFF
--- a/packages/message-resolver/src/index.ts
+++ b/packages/message-resolver/src/index.ts
@@ -1,4 +1,4 @@
-import { isObject } from '@intlify/shared'
+import { isObject, hasOwn } from '@intlify/shared'
 
 /** @VueI18nGeneral */
 export type Path = string
@@ -333,7 +333,7 @@ export function handleFlatJson(obj: unknown): unknown {
 
   for (const key in obj as object) {
     // check key
-    if (!obj.hasOwnProperty(key)) {
+    if (!hasOwn(obj, key)) {
       continue
     }
 
--- a/packages/shared/src/index.ts
+++ b/packages/shared/src/index.ts
@@ -113,6 +113,11 @@ export function escapeHtml(rawText: string): string {
     .replace(/'/g, '&apos;')
 }
 
+const hasOwnProperty = Object.prototype.hasOwnProperty
+export function hasOwn(obj: object | Array<any>, key: string): boolean {
+  return hasOwnProperty.call(obj, key)
+}
+
 /* eslint-enable */
 
 /**
--- a/packages/vue-i18n/src/composer.ts
+++ b/packages/vue-i18n/src/composer.ts
@@ -16,7 +16,8 @@ import {
   isBoolean,
   isPlainObject,
   makeSymbol,
-  isObject
+  isObject,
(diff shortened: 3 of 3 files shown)


kazupon (changes requested):
Thank you for your PR again!

in order to check to fix $305, Could you add test case ? 🙏

PeterAlfredLee:
Hi @kazupon, Just tested with this [repository](https://github.com/MartB/i18n-bug) that provided by issue #305 reporter,
and this PR will fix.

Issue #305 is caused by statement like this:
>import * as en from "./locales/en.json";

`en`  is a `Module` instead of a normal `Object` in this case.  So it will fail when we use `hasOwnProperty` call because `Module` don't have a `prototype` that point to `Object`.

I tried to add a test case but I failed.
Because in `test repository`(https://github.com/MartB/i18n-bug), the statement of `import * as en from "./locales/en.json";`  produce a `Module`, but in `vue-i18n-next` test file it produce a `Object`.

All in all, this PR can fix #305 (I have tested it), but I can't reproduce the problem when I try to add a test. WDTY ?

kazupon:
@PeterAlfredLee 
As commented in #294, I noticed that PR to be a braecking change.
https://github.com/intlify/vue-i18n-next/pull/294#issuecomment-764185615

vue-i18n-next is in RC phase right now, so I'll try to revert #294.

I think there are still a lot of things to consider to support #271.
I appreciate your contribution.

PeterAlfredLee:
> vue-i18n-next is in RC phase right now, so I'll try to revert #294.

@kazupon Sure, I understand and agree.

kazupon:
@PeterAlfredLee 
Thank you for your agreement.

In addition, We'll close also PR.
After v9 release, we’ll support #271 again.

PR #312, opened 2021-01-22
Fix spelling line 195 for sfc doc : lessages to messages

s3bc40:
I think it was "messages" unless my English is failing me. My bad if it is the case. Thanks for your work btw !


kazupon (approved):
good catch!
thx!

PR #318, opened 2021-01-26
simplify some tests

PeterAlfredLee:
Simplify some tests: 
1.replace `===` with `toEqual`
2.replace `!==` with `not.toEqual`

This makes statement shorter and more precise.


kazupon (approved):
Thanks a lot!

PR #325, opened 2021-01-27
Fix for function `deepCopy`

PeterAlfredLee:
The existing deepcopy has some potential problems : 
1. An error would be met if `destintion` is not a object.
2. And the `source[key]` with type of array would be transformed into some kind of object like `{0: array_item0, 1: array_item1, ...}`.

These problems could be tested with my testcase.

This PR has a fix for `deepcopy`:
Replace the `des[key]` with `src[key]` if eithor of them is not a object, or `src[key]` is a array(to avoid the transform from array to object I mentioned above). And recursively call `deepcopy` in other scenarios.

P.S : If `des[key]` and `src[key]` are both type of array, I would replace `des[key]` with `src[key]`, instead of combining them. I think this is a reasonable operation cause we would not expect a combination of arrays. WDYT?


kazupon (approved):
>I think this is a reasonable operation cause we would not expect a combination of arrays. WDYT? 

Yeah, I think so too.
Thanks!

PR #330, opened 2021-02-01
fix won't fallback when `fallbackLocale` is empty string, and add a test

### LINKED ISSUE #328: Fallback does not fallbacks if locale option is an empty string
### vue & vue-i18n version
- vue: 3.0.5
- vue-i18n: 9.0.0-rc.4

### Steps to reproduce

Setup i18n in this way:

```js
const i18n = {
    locale: '',
    fallbackLocale: 'en',
    messages: {/* my translations */},
};

createI18n(i18n);

createApp({
...
})
.use(i18n)
.mount('#app');

```

### What is Expected?

Fallback to 'en' locale.

Displaying messages placeholders.

PeterAlfredLee:
Fix #328

Fix won't fallback when `fallbackLocale` is empty string, and add a test.

### DIFF
--- a/packages/core-base/src/context.ts
+++ b/packages/core-base/src/context.ts
@@ -353,13 +353,9 @@ export function handleMissing<Message = string>(
 export function getLocaleChain<Message = string>(
   ctx: CoreCommonContext<Message>,
   fallback: FallbackLocale,
-  start: Locale = ''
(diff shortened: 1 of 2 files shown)


kazupon (approved):
LGTM! 👍

PR #334, opened 2021-02-02
fix for function `mergeLocaleMessage`, and modify some tests

PeterAlfredLee:
The existing `mergeLocaleMessage` use `Object.assign` to merge. 
There is some problems when merge two complex object.
These problems could be tested with my testcase which I have modified in this PR.
This PR would like to fix these problems.


kazupon (approved):
That's great works!
Thanks!

PR #345, opened 2021-02-13
fix: convert Set with Array.from

### LINKED ISSUE #344: vite:legacy built got SyntaxError: Unexpected identifier
### vue & vue-i18n version

- vue: 3.0.5
- vue-i18n: 9.0.0-rc.6

### Reproduction Link

https://github.com/sayuti-daniel/vue-i18n-repro-android-5

### Steps to reproduce

```bash
npm run build
npm run serve
```

Visit http://locahost:5000
I used https://www.ldplayer.net/:
- Android 5.1 version
- Chrome Mobile WebView version: 52.0.2743.

### What is Expected?

Page rendered correctly with translations

Blank page with `SyntaxError: Unexpected identifier` error caused by:

```js
function __msg__ (ctx) {
  const { [object Set]: _[object Set] } = ctx
  return _normalize([
    "Hello World!"
  ])
}
```

I used `vite@2.0.0-beta.67` with `@vitejs/plugin-legacy@2.0.0-beta.67` to build the project and notice [this line](https://github.com/intlify/vue-i18n-next/blob/v9.0.0-rc.6/packages/message-compiler/src/transformer.ts#L92) is the root cause which is being transpiled to:

```js
ast.helpers = [].concat(context.helpers);
```

Replacing it with `Array.from` can fix the issue.

```diff
-  ast.helpers = [...context.helpers]
+  ast.helpers = Array.from(context.helpers)
```

sayuti-daniel:
fixes #344

### DIFF
--- a/packages/message-compiler/src/transformer.ts
+++ b/packages/message-compiler/src/transformer.ts
@@ -25,7 +25,8 @@ type Transformer = Readonly<{
 }>
 
 function createTransformer(
-  ast: ResourceNode, options: TransformOptions = {} // eslint-disable-line
+  ast: ResourceNode,
+  options: TransformOptions = {} // eslint-disable-line
 ): Transformer {
   const _context = {
     ast,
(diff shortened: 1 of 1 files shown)


kazupon (approved):
Thanks!

PR #353, opened 2021-02-22
Ignore debug.log file that create by vscode

PeterAlfredLee:
Sometimes `vscode` will create a `debug.log` file, maybe we can ignore it.


kazupon (approved):
Good catch!
Thanks! :)

PR #354, opened 2021-02-22
Fix typo in breaking.md

ota-meshi:


ota-meshi:
I mistake.

ota-meshi:
Hmm. I didn't make a mistake.

ota-meshi:
We maybe also need to fix the `<i18n-n>` documentation.

ota-meshi:
I've never used the tag omission feature, but as far as I can see in the docs, I think the fix is correct.

kazupon (approved):
Thank you for you feedback!

Yeah, You're right!
The following v8.x code:
https://github.com/kazupon/vue-i18n/blob/bc4c62d7f6c8e4ebb2139d2e55e414e4db69a69b/src/components/interpolation.js#L43-L44

PR #357, opened 2021-02-23
fix: cannot apply custom pluralization rules

kazupon:
closes https://github.com/intlify/vue-i18n-loader/issues/158


MBurchard:
With this update, I got today in my project, I got the following error:

error

And I have no idea about the reason.

The project works well with rc.7 but not with rc.8

PR #362, opened 2021-02-24
update example project vite version to ^2.0.2

PeterAlfredLee:
The official version of vite 2.0.0 has been released. 
I think we can update the vite version of the example project to the latest version (2.0.2).

I've just tested it on my machine, we just need a little change to the `i18n. ts` file, then everything works fine.


kazupon (approved):
Thanks!

PR #365, opened 2021-02-26
fix: undefind setter using

kazupon:


kazupon:
resove https://github.com/intlify/vue-i18n-next/pull/357#issuecomment-786515810

PR #369, opened 2021-03-01
Fix eslint warning in example project

PeterAlfredLee:
There is a warning in example project `examples/lazy-loading/webpack` and `examples/lazy-loading/vite` :
```
[vue/valid-template-root]
The template root requires exactly one element.eslint-plugin-vue
```

This PR is a fix for it.


kazupon (approved):
Thanks!

Issue #395, opened 2021-03-13
docs: improve component interpolation for pluralization


kazupon:
We need to add the pluralization section
https://vue-i18n.intlify.dev/guide/advanced/component.html

ref: https://github.com/intlify/vue-i18n-next/discussions/394

welcome contribution :)

PR #398, opened 2021-03-16
docs: add pluralization usage section in component interpolation section

PeterAlfredLee:
This PR would like to add the pluralization section in component interpolation section.

Relate to #395


kazupon (approved):
Thanks always!

PR #402, opened 2021-03-18
Add support for flat json separated with dot('.')

PeterAlfredLee:
Relate to issue #271. Have done in #294, and then we decided this should wait after v9 release.  See this [comment](https://github.com/intlify/vue-i18n-next/pull/306#issuecomment-764208669).
Now v9 have released, I think it's a good time to add this feature.

Note, this time I make this feature optional and not enabled by default.
So I add option `flatJson` in some `interface`. 
Hope I did them in a right way ^_^.

Here is what I have done:
1. add function `handleFlatJson` to `message-resolver`,
function `handleFlatJson` is use to transform flat json to normal json.

2. add optional support for flat json messages to `vue-i18n`.

3. add some tests to `massage-resolver` and `vue-i18n`.

4. move function `hasOwn` to `@intlify/shared`.

### DIFF
--- a/packages/message-resolver/src/index.ts
+++ b/packages/message-resolver/src/index.ts
@@ -1,4 +1,4 @@
-import { isObject } from '@intlify/shared'
+import { isObject, hasOwn } from '@intlify/shared'
 
 /** @VueI18nGeneral */
 export type Path = string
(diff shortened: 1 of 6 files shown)


kazupon (approved):
That's great works!
Thanks!

PR #404, opened 2021-03-19
fix: suppress console for devtools flag

kazupon:


kazupon:
ref: https://github.com/intlify/vite-plugin-vue-i18n/issues/74#issuecomment-802665616

PR #410, opened 2021-03-27
fix warning: Missing associated label

PeterAlfredLee:
There is a warning "`Missing associated label` " in some vue file and html file. 
This PR is a fix for it.
See also this [MDN page](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select)


kazupon (approved):
Thanks a lot!

PR #418, opened 2021-04-01
Modify some router setups in example projects

PeterAlfredLee:
Modify navigation guards:

1. Remove parameter `from` because it's not used
2. Remove parameter `next`， use recommended way by `vue router ` for instead, see also [this](https://next.router.vuejs.org/guide/advanced/navigation-guards.html#global-before-guards)


kazupon (approved):
Thanks!
This fix is for example code.
so, I'll merge it! :)

PR #453, opened 2021-04-13
Support passing composer instance to Translation component

andresberrios:
When using `<i18n-t>` inside a slot of a custom component, it will try to use the scope of the parent component, but because of being inserted into a slot, its parent is now that custom component instead of the original parent component, where the template that used `<i18n-t>` was compiled. As a more concrete example:

1. `Parent` component defines an i18n scope, and a message with key `hello`.
2. `Parent` component uses `Child` component in its template.
3. `Child` component defines a slot named `content` and a new i18n scope (without any messages).
4. `Parent` component uses `<i18n-t keypath="hello">` inside the `content` slot of `Child` component.
5. `<i18n-t>` looks for the message with key `hello` but can't find it in its "parent" component because it's not longer `Parent`, but `Child` instead.

The expected behavior would be for `<i18n-t>` to look for the messages in `Parent` component instead, since that's where it was used in the first place.

I've looked through the source code and can't find a way to somehow trace the parent lookup chain back to where the Translation component was used instead of where it lands through the slot.

The only way I found to make this work was to allow for passing an actual `i18n` composer instance to the `<i18n-t>` component as a prop, so it can be directly used if provided, or fall back to the regular behavior otherwise. This way we can pass the `i18n` composer instance from `Parent` component directly to `<i18n-t>` even if it's placed inside the slot of a custom component that creates its own i18n scope.

Hopefully you find this useful and decide to merge it!

### DIFF
--- a/packages/vue-i18n/src/components/Translation.ts
+++ b/packages/vue-i18n/src/components/Translation.ts
@@ -27,6 +27,11 @@ export interface TranslationProps extends BaseFormatProps {
    * The Plural Choosing the message number prop
    */
   plural?: number | string
+  /**
+   * @remarks
+   * An existing i18n Composer instance to use for translating
+   */
+  i18n?: Composer
 }
 
 /**
--- a/packages/vue-i18n/src/components/Translation.ts
+++ b/packages/vue-i18n/src/components/Translation.ts
@@ -91,8 +91,7 @@ export const Translation = {
         type: [Number, String],
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
         validator: (val: any): boolean => isNumber(val) || !isNaN(val)
-      },
-      i18n: { type: Object }
+      }
     },
     baseFormatProps
   ),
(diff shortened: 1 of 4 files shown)


packages/vue-i18n/src/components/Translation.ts
```diff
@@ -27,6 +27,11 @@ export interface TranslationProps extends BaseFormatProps {
    * The Plural Choosing the message number prop
    */
   plural?: number | string
+  /**
+   * @remarks
+   * An existing i18n Composer instance to use for translating
+   */
+  i18n?: Composer
```
kazupon, line 34: I think this issue is not only with the `Translation` component but also with the `DatetimeFormat` and `NumberFormat` components.
 
I hope that you could define it to `BaseFormatProps` interface and `baseFormatProps` prop. 
https://github.com/intlify/vue-i18n-next/blob/master/packages/vue-i18n/src/components/base.ts

kazupon (changes requested):

packages/vue-i18n/src/components/Translation.ts
andresberrios: Done! 😃

kazupon (approved):

kazupon:
good job! 👍

andresberrios:
@kazupon thank you for all the amazing work! Hontoni arigato gozaimasu

PR #480, opened 2021-04-26
Simplify some condition judgments

PeterAlfredLee:
Minor improvements for some condition judgment to make them more clear.


kazupon (approved):
Thanks a lot!

Issue #481, opened 2021-04-26
$n ignores formatting options


johannesstricker:
`$n` seems to ignore formatting options.
```javascript
$n(1.234234, { maximumFractionDigits: 2 }); // -> returns 1.234 instead of 1.23
```

**Module versions (please complete the following information):**
 - `vue`: 3.0.11
 - `vue-i18n`: 9.1.6

**Reproduction Link**
https://codesandbox.io/s/vue-i18n-9-template-forked-45g0k

**Expected behavior**
Since `maximumFractionDigits: 2` is specified, the number should be formatted with 2 decimals. Instead it shows the default 3 decimals.

PeterAlfredLee:
Hi, @johannesstricker 

This might be a configuration error. See this Link: https://codesandbox.io/s/vue-i18n-9-template-forked-2gxxq
If you add these config
```js
numberFormats: {
  en: {
	decimal: {
	  maximumFractionDigits: 2
	}
  },
  ja: {
	decimal: {
	  maximumFractionDigits: 2
	}
  }
}
```
and call `$n` like this : 
```html
<p>{{ $n(1.123123, "decimal") }}</p>
```
everything works fine.

Also, if you want to overwrite the `maximumFractionDigits`, you can call `$n` like this:
```html
<p>{{ $n(1.123123, "decimal", {maximumFractionDigits: 5}) }}</p>
```

Hope this help.

johannesstricker:
Thank you, that did work 👍  I was really confused because there was no need to specify the number format in v8 and the docs are not really explicit about this.

PR #487, opened 2021-04-29
Add some tests about numberFormats

PeterAlfredLee:
Some users may not be familiar with `numberFormats`. See #481

This PR consists of some tests about `minimumSignificantDigits, maximumSignificantDigits` and `notation`. I'm trying to show some examples about the usage of these configurations. Hoping this can make it more clear.

### DIFF
--- a/docs/guide/essentials/number.md
+++ b/docs/guide/essentials/number.md
@@ -10,12 +10,24 @@ Number formats the below:
 const numberFormats = {
   'en-US': {
     currency: {
-      style: 'currency', currency: 'USD'
+      style: 'currency', currency: 'USD', notation: 'standard'
+    },
+    decimal: {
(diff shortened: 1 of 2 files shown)


kazupon (commented):
Thank you for your feedback!

I think that issue was opened because the docs example for NumberFormat was not good docs.
So, we should be added more example cases to NumberFormat section on vue-i18n docs.

What do you think about my comment?

PeterAlfredLee:
> So, we should be added more example cases to NumberFormat section on vue-i18n docs.

Agreed. I just pushed a new commit about it.
Feel free to let me know if there are something else to be modified. :-)

kazupon (approved):

kazupon:
Thank you! 
You're always so reliable. :)

PeterAlfredLee:
> You're always so reliable. :)

I'm glad this helps a little bit. :-)

PR #544, opened 2021-06-18
fix: wrong vue-i18n path config in some example projects

PeterAlfredLee:
In some example projects, It seems that the relative path of vue-i18n is wrong.
For example, in `examples/lazy-loading/vite/package.json`, Now the configuration is :
```
"vue-i18n": "link:../../packages/vue-i18n"
```
But the correct relative path would be
```
"vue-i18n": "link:../../../packages/vue-i18n"
``` 

This PR fix these wrong conifgs, and update some yarn.lock files.

### DIFF
--- a/examples/lazy-loading/vite/package.json
+++ b/examples/lazy-loading/vite/package.json
@@ -7,7 +7,7 @@
   },
   "dependencies": {
     "vue": "^3.0.11",
-    "vue-i18n": "link:../../packages/vue-i18n",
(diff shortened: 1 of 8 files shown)


PeterAlfredLee:
Just added a commit to fix lint error of file `packages/vue-i18n/src/vue.d.ts`.
This lint error is not caused by this PR.

kazupon (approved):
Thanks!

PR #606, opened 2021-08-03
Fix type-safe example warning

PeterAlfredLee:
1. Fix `not found key` warning in `examples/type-safe/global-type-definition`,   see [this implementation](https://github.com/intlify/vue-i18n-next/blob/6eaf91ae864f128676d5f6000986ee8ae31f77cf/packages/vue-i18n-core/src/i18n.ts#L599)
```ts
  const scope: I18nScope = isEmptyObject(options)
    ? ('__i18n' in instance.type)
      ? 'local'
      : 'global'
    : !options.useScope
      ? 'local'
      : options.useScope
```
It will set scope to `local` when `options` is not empty and `options.useScope` was not config. So in this case we should explicit use global scope.

2. Fix  `feature flag` warning in `examples/type-safe/global-type-definition` and `examples/type-safe/type-annotation` by config flags in `vite.config.ts`

### DIFF
--- a/examples/type-safe/global-type-definition/src/components/HelloWorld.vue
+++ b/examples/type-safe/global-type-definition/src/components/HelloWorld.vue
@@ -13,6 +13,7 @@ export default defineComponent({
   setup() {
     // use global scope
     const { t, d, n } = useI18n({
+      useScope: 'global',
       inheritLocale: true
     })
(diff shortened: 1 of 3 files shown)


kazupon (approved):
Good catch!
Thanks!

PR #664, opened 2021-09-06
Fixed a type in the Composition API documentation (it's -> its)

SeverityOne:
In the page about the Composition API, there's an "it's" that should be spelled "its" instead. This was fixed.


kazupon (approved):
Thanks!

PR #666, opened 2021-09-13
fix: source-map bundling on message-compiler esm bundler

kazupon:
revert #619


kazupon:
merge, because CI passing on PR

PR #742, opened 2021-10-14
fix: node esm for nuxt3

kazupon:


li9269391:
How to use it in nuxt3？

Issue #746, opened 2021-10-15
Runtime Error - TypeError: o is not a function


jiaozhiye:
### Reporting a bug?

Update to 9.2.0-beta.12, js error, no problem with previous beta.

### Expected behavior

I hope there are no errors

### Reproduction

An error occurs when you update to this version

### System Info

```shell
System:
    OS: Windows 10 10.0.19043
    CPU: (8) x64 Intel(R) Core(TM) i7-8565U CPU @ 1.80GHz
    Memory: 8.43 GB / 15.81 GB
  Binaries:
    Node: 14.16.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.11 - C:\Program Files\nodejs\yarn.CMD
    npm: 6.14.11 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1023.0), Chromium (94.0.992.47)
    Internet Explorer: 11.0.19041.1202
  npmPackages:
    @vue/babel-plugin-jsx: ^1.1.0 => 1.1.0
    @vue/compiler-sfc: ^3.2.20 => 3.2.20
    vue: ^3.2.20 => 3.2.20
    vue-eslint-parser: ^7.6.0 => 7.6.0
    vue-i18n: ^9.2.0-beta.12 => 9.2.0-beta.12
    vue-loader: ^16.8.1 => 16.8.1
    vue-router: ^4.0.11 => 4.0.11
    vue-style-loader: ^4.1.3 => 4.1.3
    vuex: ^4.0.2 => 4.0.2
```

### Screenshot

![image](https://user-images.githubusercontent.com/22720488/137439275-a557bbf2-4d50-4fd0-bf92-a65494543790.png)

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/intlify/vue-i18n-next/blob/master/.github/CONTRIBUTING.md)
- [X] Read the [Documentation](https://vue-i18n.intlify.dev/)
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/intlify/vue-i18n-next/discussions)

jiaozhiye:
> I also have this problem
> 
> @jiaozhiye you can use "vue-i18n": "9.2.0-beta.11",

Yes, 9.2.0-Beta.11 does not have this problem

zeyadsharo:
[...]

PR #748, opened 2021-10-19
fix not a function error

PeterAlfredLee:
See this [comment](https://github.com/intlify/vue-i18n-next/issues/746#issuecomment-946286908) in #746


kazupon (approved):
Thank you for your good work!

The package that is part of named `runtime` doesn't include the message compiler.
So, This fix makes sense! :)

Issue #776, opened 2021-11-08
Cannot silence fallback warnings like on V8


LifeIsStrange:
### Reporting a bug

By default when fallbacking fail, two warnings are printed
e.g
```
[intlify] Fall back to translate 'bf.headerBar.homeSearchBarPlaceholder' key with 'en' locale
[intlify] Not found 'bf.headerBar.homeSearchBarPlaceholder' key in 'en' locale messages.
```
The [documentation](https://vue-i18n.intlify.dev/guide/essentials/fallback.html#explicit-fallback-with-one-locale) clearly state:
> To suppress these warnings (**while keeping those which warn of the total absence of translation for the given key**) set silentTranslationWarn: true, and silentFallbackWarn: true when initializing the createI18n.

Since the failing fallback warnings are **much** too noisy, I want to silence them, but **keep** other i18n warnings.
Unfortunately, it seems that those booleans have been *silently* renamed and the documentation no longer reflect the state of V9
Apparently silentFallbackWarn has become fallbackWarn and the boolean logic reversed. If so why not I don't care and indeed setting fallbackWarn remove **half** of the warnings.
silentTranslationWarn seems to now longer exist ? (at least in non-legacy mode), however I cannot find the new boolean.
It should have been renamed to translationWarn to be consistent with fallbackWarn renaming but there is no such thing.
The ComposerOptions type expose the following candidates:
* missingWarn: when enabled it effectively hide warnings, but it hide too much warnings e.g I have no key in the fallback languages nor in the user locale. 
This is in direct contradiction with 
> while keeping those which warn of the total absence of translation for the given key

* missing: this boolean might be the new silentTranslationWarn ? weird cause the type is not a boolean despite what the doc claims, I have no idea how to set missing.

I am on the latest stable Vue-i18n 9.1.x release.

``` Typescript
const i18n = createI18n({
  fallbackWarn: false,
  missingWarn: false,
  fallbackLocale: ["en", "fr"],
  globalInjection: true,
  legacy: false,
  useScope: "global",
  locale: locale,
  messages: loadTranslations(),
  datetimeFormats,
  numberFormats,
});
```
edit:
the doc confirm that silentTranslationWarn has been [renamed](https://vue-i18n.intlify.dev/guide/advanced/composition.html#mapping-between-vuei18n-instance-and-composer-instance
) to missingWarn.
Then how come I have now warning when no key is found for fallbackLocale?
@kazupon friendly ping

### Validations

[...]

PR #778, opened 2021-11-09
docs: improve doc of fall back warinings

PeterAlfredLee:
See also #776

### DIFF
--- a/docs/guide/essentials/fallback.md
+++ b/docs/guide/essentials/fallback.md
@@ -62,7 +62,9 @@ The first warning message is printed the key, due to  given to the translation f
 These warning messages are only warned in development mode (`process.env.NODE_ENV !== 'production'`) by default, not for production.
 :::
 
-To suppress these warnings (while keeping those which warn of the total absence of translation for the given key) set `silentTranslationWarn: true`, and `silentFallbackWarn: true` when initializing the `createI18n`.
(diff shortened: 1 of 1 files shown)


kazupon (approved):
Thanks!

LifeIsStrange:
I think (while keeping those which warn of the total absence of translation for the given key) should still be stated. Otherwise, thanks for the clarification!

JWong1105:
Hello everybody, just now I found the right config by debug the source code.
We just can only set 'silentTranslationWarn: true' and 'silentFallbackWarn: true' 
the source codes are in the file 'vue-i18n.esm-bundler.js' at line 661.
`    const missingWarn = isBoolean(options.silentTranslationWarn) ||
        isRegExp(options.silentTranslationWarn)
        ? !options.silentTranslationWarn
        : true;
    const fallbackWarn = isBoolean(options.silentFallbackWarn) ||
        isRegExp(options.silentFallbackWarn)
        ? !options.silentFallbackWarn
        : true;`

JWong1105:
and the 'warn codes' are at line 347.
`                if (_fallbackRoot &&
                    (isTranslateFallbackWarn(_fallbackWarn, key) ||
                        isTranslateMissingWarn(_missingWarn, key))) {
                    warn(getWarnMessage(6 /* FALLBACK_TO_ROOT */, {
                        key,
                        type: warnType
                    }));
                }`

PR #784, opened 2021-11-11
fix(core-base): fallback interpolation should not break in runtime only (fix #768)

### LINKED ISSUE #768: Runtime only build breaks with missing keys despite formatFallback
### Reporting a bug?

Hello, recently with my team we decided to optimize vue-i18n following your guide recommendations and ended up using the official vite plugin. 
Now we run into an issue with runtimeOnly build where missing message key are making the app crash, despite `formatFallback: true` because it tries to compile the fallback message key as a string.

### Expected behavior

with `formatFallback: true` I expect that missing message keys are not hard failing.
That is a show stopper for use since our translation process is asynchronous and missing translations are common. For now, we are still including the message compiler in our build only because of this issue.

A possible fix would be to convert `key: string` to  `() => key`   if `formatFallback && !messageCompiler`

### Reproduction

https://github.com/AlexandreBonaventure/vue-i18n-runtime-only

### System Info

```shell
irrelevant
```

### Screenshot

Screen Shot 2021-11-04 at 12 13 02 PM

### Additional context

Thanks!

### Validations

- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/intlify/vue-i18n-next/discussions)

AlexandreBonaventure:
Fix #768

### DIFF
--- a/packages/core-base/src/translate.ts
+++ b/packages/core-base/src/translate.ts
@@ -342,6 +342,7 @@ export function translate<
     fallbackFormat,
     postTranslation,
     unresolving,
+    messageCompiler,
     fallbackLocale,
     messages
   } = context
@@ -368,7 +369,7 @@ export function translate<
         ? options.default
         : key
       : fallbackFormat // default by `fallbackFormat` option
-        ? key
+        ? (!messageCompiler ? () => key : key)
         : ''
   const enableDefaultMsg = fallbackFormat || defaultMsgOrKey !== ''
   const locale = isString(options.locale) ? options.locale : context.locale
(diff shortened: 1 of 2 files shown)


kazupon (changes requested):
And Welcome contribution! ❤️ 

I've just checked your modifying codes
It's OK!
That's great work!

We're getting an error in lint, so Could you please fix that? 🙏

AlexandreBonaventure:
Ok sorry eslint is not working out-of-the box for me with VSCode. I get a dependency resolving issue despite running `pnpm install`.
Screen Shot 2021-11-15 at 10 57 02 AM

Nevertheless, I fixed it now, you can review again, thanks!

kazupon (approved):

PR #848, opened 2021-12-27
fix typo petitle-vue-i18n

ElenaTv:


kazupon (approved):
Good catch!
Thank you very much!

PR #897, opened 2022-01-24
improve types declaration

### LINKED ISSUE #889: [typescript] I18n.global incorrect type when in "composition" mode (should be Composer, is VueI18n)
### Reporting a bug?

This is a TypeScript-related bug.

From the docs https://vue-i18n.intlify.dev/api/general.html#global it is said that "If the I18n#mode is 'legacy', then `I18n.global` can access to a global VueI18n instance, else then I18n#mode is 'composition' , you can access to the global Composer instance.". 

I've found this to be true at runtime, but the inferred TypeScript type is always `VueI18n` even when in "composition" mode. This throws errors at compile time and in the editor when I try to access Composer properties which are differently typed or absent from VueI18n (according to the mapping found at https://vue-i18n.intlify.dev/guide/advanced/composition.html#mapping-between-vuei18n-instance-and-composer-instance). More on that, this breaks typing on functions that return `useI18n() | i18n.global`.

### Expected behavior

The following code should not throw a typescript error on the last line 

```ts
import { createI18n } from "vue-i18n";

const testI18n = createI18n({
  legacy: false,
  locale: "ja",
  fallbackLocale: "en",
  messages: {
    en: {
      message: {
        hello: "hello world",
      },
    },
    ja: {
      message: {
        hello: "こんにちは、世界",
      },
    },
  },
});

const globalComposerInstance = testI18n.global;
console.log(globalComposerInstance.formatter); // this should be absent in Composer instance!
console.log(globalComposerInstance.getMissingHandler); // this should be present is Composer instance!
```

Instead it throws the following error:
```
Property 'getMissingHandler' does not exist on type 'VueI18n<{ en: { message: { hello: string; }; }; ja: { message: { hello: string; }; }; }, {}, {}, string, \"ja\" | \"en\", \"ja\" | \"en\", Composer<{ en: { message: { hello: string; }; }; ja: { message: { hello: string; }; }; }, ... 4 more ..., \"ja\" | \"en\">>'.
```

### Reproduction

Use the above snippet anywhere in a project with `vue-i18n@9.2.0-beta.28` and `typescript@4.5.4`. Hope it is enough, if not i will provide a repro.

### System Info

```shell
System:
    OS: Linux 5.11 Linux Mint 20.2 (Uma)
    CPU: (12) x64 Intel(R) Core(TM) i7-10710U CPU @ 1.10GHz
    Memory: 843.50 MB / 15.47 GB
    Container: Yes
    Shell: 5.8 - /usr/bin/zsh
  Binaries:
    Node: 14.15.0 - ~/.nvm/versions/node/v14.15.0/bin/node
    Yarn: 1.22.17 - ~/.nvm/versions/node/v14.15.0/bin/yarn
    npm: 8.1.0 - ~/.nvm/versions/node/v14.15.0/bin/npm
  Browsers:
    Brave Browser: 97.1.34.80
    Chrome: 97.0.4692.71
    Firefox: 95.0.1
  npmPackages:
    vue-i18n: ^9.2.0-beta.28 => 9.2.0-beta.28
```

### Screenshot

immagine

### Additional context

The actual value of `testI18n.global` is in fact a Composer instance. The following code:

```ts
console.log("testI18n.global", testI18n.global);
console.log("testI18n.global.formatter", testI18n.global.formatter);
```

logs the following: 

immagine

### Validations

- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/intlify/vue-i18n-next/discussions)
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

littleboarx:
resolve #889

### DIFF
--- a/packages/vue-i18n-core/src/i18n.ts
+++ b/packages/vue-i18n-core/src/i18n.ts
@@ -311,7 +311,11 @@ export function createI18n<
 >(
   options: Options,
   LegacyVueI18n?: any // eslint-disable-line @typescript-eslint/no-explicit-any
-): I18n<Messages, DateTimeFormats, NumberFormats, OptionLocale, Legacy>
+): typeof options['legacy'] extends true
+  ? I18n<Messages, DateTimeFormats, NumberFormats, OptionLocale, true>
+  : typeof options['legacy'] extends false
+  ? I18n<Messages, DateTimeFormats, NumberFormats, OptionLocale, false>
+  : I18n<Messages, DateTimeFormats, NumberFormats, OptionLocale, Legacy>
 
 /**
  * Vue I18n factory
@@ -410,7 +414,11 @@ export function createI18n<
 >(
   options: Options,
   LegacyVueI18n?: any // eslint-disable-line @typescript-eslint/no-explicit-any
-): I18n<Messages, DateTimeFormats, NumberFormats, OptionLocale, Legacy>
+): typeof options['legacy'] extends true
+  ? I18n<Messages, DateTimeFormats, NumberFormats, OptionLocale, true>
+  : typeof options['legacy'] extends false
+    ? I18n<Messages, DateTimeFormats, NumberFormats, OptionLocale, false>
+    : I18n<Messages, DateTimeFormats, NumberFormats, OptionLocale, Legacy>
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/explicit-module-boundary-types
 export function createI18n(options: any = {}, VueI18nLegacy?: any): any {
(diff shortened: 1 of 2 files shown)


kazupon (changes requested):
Welcome!

ESLint errors have occurred in CI
Could you fix these errors, please?

kazupon (approved):

kazupon:
Thanks!
Good first contribution!
I'll be merging now!

PeterAlfredLee:
Hi, @littleboarx , I just checked this PR.
I think the type of `strictI18n` should be `Composer<>` instead of `Vuei18n<>`. WDYT?
image

littleboarx:
> Hi, @littleboarx, I just checked this PR. I think the type of `strictI18n` should be `Composer<>` instead of `Vuei18n<>`. WDYT? image

I found out that if you use the generic parameter, it will use the default generic value instant of inferring from the function call. and I edit the test case in line 399. The change to this test case needs to be rolled back.

kazupon:
@littleboarx 
If you can resolve it, you can send PR as a fix, else you can't it, we need to roll back.

littleboarx:
see this PR:
https://github.com/intlify/vue-i18n-next/pull/903

> @littleboarx If you can resolve it, you can send PR as a fix, else you can't it, we need to roll back.

PR #903, opened 2022-01-25
fix: revert type test case change in strictI18n

littleboarx:
fix: revert type test case change in strictI18n


kazupon (approved):
Thanks!

PR #960, opened 2022-03-17
docs: update process.env.NODE_ENV

YunYouJun:
It seems we do not need to escape `\.` in `<code></code`.


kazupon:

I already understand this issue. And there is already an issue.
https://github.com/vuejs/vitepress/issues/528

The reason why I am doing this is to avoid some issues.
We can work around it by bringing the vitepress version up to date, but the vue-i18n api docs will not build. api docs are generated by `@microsoft/api-extractor`, but the generated markdowns are unfortunately It seems to be incompatible with vitepress 😞 

I appreciate your contribution!

PR #1035, opened 2022-06-22
fix: fix "pnpm dev" failure

lazzzis:
Per [contribution guide](https://github.com/intlify/vue-i18n-next/blob/e94386aa6b0999991582bc2051673314360d14a0/.github/CONTRIBUTING.md?plain=1#L135), running `pnpm dev` should build vue-i18n package, but it actually causes a failure:

```bash
> node -r esbuild-register scripts/dev.ts


   ERROR  Target  not found!

 ELIFECYCLE  Command failed with exit code 1.
FAIL: 1
```

This seems because fuzzyMatchTarget throws an error when no command line argument is passed in.


kazupon:
Thanks!

PR #1039, opened 2022-06-23
fix: typo in the layout slot

xfq:
Fix a typo in a slot of the default theme's `<Layout/>` component.


kazupon (approved):
Good catch!
Thanks!

PR #1040, opened 2022-06-25
fix: fix isExist always returns false and polish build.ts

lazzzis:
1. fix that `isExist` always returns false no matter if the file exists
2. fix `await` that has no effect in line 205: `await !isExist(filePath)` => `!(await isExist(filePath))`
3. polish code and try to add some types btw

Thanks!

### DIFF
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -69,13 +69,17 @@ const rimraf = promisify(_rimraf)
     }
   }
 
-  async function buildAll(targets) {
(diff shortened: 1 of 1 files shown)


kazupon (approved):
Thank you for your contribution! ❤️ 

I’m happy about your contribution. your contribution is an improvement for our dev ops of the project. 
We have some developments for our own projects.
We are really happy to have contributors like you.

Thanks!

PR #1041, opened 2022-07-01
replace docs site with vitepress v1

kazupon:


xfq:
Since VitePress v1 supports dark mode, you may want to add something like:

```css
.dark .banner {
  filter: grayscale(1) invert(1);
}
```

so that the sponsor logos will look better in dark mode.

BTW, it seems that you removed the links to the Japanese documentation, it would be great if they can be brought back in the future.

kazupon:
@xfq 
Thank you for your advise.

About the sponsor, I've just updated with vitepress sponsor component.
so, I could fix it.



> BTW, it seems that you removed the links to the Japanese documentation, it would be great if they can be brought back in the future.

Unfortunately, vitepress v1 alpha version is not supported
https://vitepress.vuejs.org/guide/migration-from-vitepress-0.html#app-config

So, I will back up japanese docs in repo until internationalization supporting

PR #1046, opened 2022-07-06
docs: add missing imports and fix typescript example of vue-i18n-bridge

kingyue737:
- Add missing `import VueI18n from 'vue-i18n'` vue-i18n-bridge examples
- Change position of `castToVueI18n` to avoid type error in `Vue.use(i18n)`


kazupon (approved):
Thanks!

PR #1047, opened 2022-07-08
Fix the id of the app's container element

xfq:
https://vue-i18n.intlify.dev/guide/advanced/directive.html#object-syntax

In the JavaScript it's `app.mount('#object-syntax')`, but in the template it's `<div id="app">`. The two do not match.


kazupon (approved):
Thanks!

PR #1064, opened 2022-07-16
fix: slot resolving in render function

kazupon:
We change the Slots keys to be resolved in the render function, which in Vue 3 can be resolved in the setup function, but in Vue 2.7 can only be resolved in the render function

related #1062


kazupon:
related issue  https://github.com/vuejs/vue/issues/12672

PR #1084, opened 2022-07-23
Added pluralization example to custom directive page

anatolykopyl:
The docs don't explain that the object syntax of the custom directive can accept the `plural` parameter.


kazupon (approved):
Thanks!

PR #1105, opened 2022-08-04
i18n.global.locale is a ref

Cygnut:
i18n.global.locale is a ref - so the example should be assigning to it in the correct manner


kazupon (approved):
Thanks!

Cygnut:
Thank you :)

PR #1114, opened 2022-08-08
fix: rename file extension of ESM modules

jaulz:
Since the `vue-i18n` package is not declared as a module the imported files are not recognized as modules. Thus, it leads to errors because the `import` statement is not allowed in CommonJS files. This PR will rename the file extensions so NodeJS loads the files correctly as modules. 

This is a breaking change because all the documents refer to a file ending with `.js` so I would like to hear your opinion what you think.

### DIFF
--- a/packages/vue-i18n/package.json
+++ b/packages/vue-i18n/package.json
@@ -53,10 +53,10 @@
   "buildOptions": {
     "name": "VueI18n",
     "formats": [
-      "esm-bundler.mjs",
-      "esm-bundler-runtime.mjs",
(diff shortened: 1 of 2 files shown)


kazupon:
Sorry that we couldn’t include your PR.

I'm working on node.js esm file extension, so we want to resolve the below issue.
https://github.com/intlify/vue-i18n-next/pull/1137

The reason is that Vite 3 and its eco system require an ESM with `mjs`.
the below issue:
https://github.com/intlify/vue-i18n-next/issues/1131

We appreciate your PR.

PR #1129, opened 2022-08-24
Add information about the pluralization option for the Composition API

domoratskii:


domoratskii:
@kazupon check pls

docs/guide/essentials/pluralization.md
```diff
@@ -141,7 +141,7 @@ function customRule(choice, choicesLength, orgRule) {
 }
 ```
 
-To use the custom rules defined above, you need to set the `pluralizationRules` option of `createI18n` to a key like the following locale:
+To use the custom rules defined above, you need to set the `pluralizationRules` option of `createI18n` if you use the Options API or the `pluralRules` if you use the Composition API to a key like the following locale:
```
prostko, line 144: Great change, thank you for pointing this out because it helped me with using this library.

However, this is a bit awkward to read in English. I would recommend changing to the following. 

---

To use the custom rules defined above, inside of `createI18n` set either: 
1. `pluralizationRules` (for Options API)
 _or_
2. `pluralRules` (for Composition API)

like the the following locale:

kazupon:
Thanks! Welcome, contribution!

PR #1130, opened 2022-08-24
fix(package): allow vue 2.7

dargmuesli:
This library should be usable with Vue 2.7 as well.

Missing: `vue.Fragment`, `vue.Text`, `vue.createVNode` support. My question:
- could those usages be checked to only run in Vue 3?
- should one use i18n-bridge instead?

### DIFF
--- a/packages/vue-i18n/package.json
+++ b/packages/vue-i18n/package.json
@@ -45,7 +45,7 @@
     "@intlify/devtools-if": "9.2.2"
   },
   "peerDependencies": {
-    "vue": "^3.0.0"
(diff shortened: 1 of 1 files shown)


kazupon:
Thank you for your feedback!

> * could those usages be checked to only run in Vue 3?
> * should one use i18n-bridge instead?

As our answer for these questions, you can use `vue-i18n-bridge`.
you can check the docs:
https://vue-i18n.intlify.dev/guide/migration/vue2.html#vue-i18n-bridge

kazupon:
we already have a solution, so lose it.
Thanks!

PR #1204, opened 2022-10-20
Updated directive.md

sayan99614:
`$t` is function of "Vue18n" instance. It has the following pros and cons.
Incorrect : "Vue18n"
Change: "VueI18n"


kazupon (approved):
Thanks!

PR #1236, opened 2022-11-18
fix: correct typo in `TranslateVNodeSymbol`

DamianGlowala:
I'm not entirely sure whether the error shown is due to the typos, but nevertheless worth correcting :)

(occured when using the `<i18n-t>` component with `@nuxtjs/i18n@next`)

image


kazupon (approved):
Welcome, contribution! :)
Good catch!
Thanks!

PR #1241, opened 2022-11-21
fix: avoid VueI18n extending on root

kazupon:


kazupon:
related: https://github.com/nuxt-community/i18n-module/issues/1659

PR #1245, opened 2022-11-22
Docs: search for "useI18n" yields non-existing /ja/api/composition.html 

Jinyoung-s:
There is a search issue on the doc site (https://vue-i18n.intlify.dev/) 
The non-English page is excluded by adding the configuration parameter of Algolia.

Issue 
https://github.com/intlify/vue-i18n-next/issues/1218


kazupon (approved):
Thank you so much!

Issue #1191, opened 2022-09-30
The requested module does not provide an export named 'default'


TheInvoker:
### Reporting a bug?

In my vite + vue 3 project, I import i18n like this

    import VueI18n from "vue-i18n";

But I then get this error in the dev tools console:

    Uncaught SyntaxError: The requested module '/node_modules/.vite/deps/vue-i18n.js?v=3de9cdf7' does not provide an export named 'default' (at i18n.ts:1:8)

How can I resovle this?

### Expected behavior

It should not throw any syntax error.

### Reproduction

No reproduction link is needed.

### System Info

```shell
System:
    OS: Windows 10 10.0.22000
    CPU: (4) x64 Intel(R) Xeon(R) Platinum 8171M CPU @ 2.60GHz
    Memory: 6.13 GB / 16.00 GB
  Binaries:
    Node: 16.13.2 - C:\Program Files\nodejs\node.EXE
    npm: 8.1.2 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.22000.120.0), Chromium (105.0.1343.53)
    Internet Explorer: 11.0.22000.120

vite 3.1.0
i18n 9.2.2
vue 3.2.37
```

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/intlify/vue-i18n-next/blob/master/.github/CONTRIBUTING.md)
- [X] Read the [Documentation](https://vue-i18n.intlify.dev/)
- [X] Check that there isn't [already an issue](https://github.com/intlify/vue-i18n-next/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/intlify/vue-i18n-next/discussions)
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

kazupon:
Thank you for your reporting!

In vue 3, vue-i18n no longer default exports. You will need to use an API like `createI18n` with named export.

Please read docs:
https://vue-i18n.intlify.dev/

Thanks!

TheInvoker:
Actually it doesn't work with vite unless you import it like `import * as VueI18n from "vue-i18n";`. That fixed it for me.

tobegit3hub:
This may work to use `createI18n` directly without `VueI18n`.

```
import { createI18n } from 'vue-i18n'

const i18n = createI18n({
[...]

PR #1301, opened 2023-01-05
Update JavaScript instructions

iuricmp:
These simple changes can avoid small mistakes like this #1191  
It also stays in sync with the sources we see in the `examples` folder.


kazupon:
Thanks for your doc improvement contribution.

Currently, the vue-i18n docs are described based for beginning javascript users.
So we are using the script tag to get started.
We trust that those familiar with javascript will be able to read the docs and understand along with the examples code.

Thanks!

Issue #1315, opened 2023-01-20
Locale returned by useI18n is incorrect when legacy mode is enabled


Csszabi98:
### Reporting a bug?

When calling `usei18n` from vue setup and using the legacy API (`legacy: true`) with `allowComposition: true`, the returned locale on the first render is always the default one ("en-US").

### Expected behavior

The returned locale should be the globally set one on the first render if the `useI18n` options contain no override.

### Reproduction

```
import { createApp } from 'vue'
// import the root component App from a single-file component.
import App from './App.vue'

const app = createApp(App)

const DEFAULT_LOCALE = 'de'

// Create i18n instance
const i18n = createI18n({
    allowComposition: true,
    fallbackLocale: DEFAULT_LOCALE,
    legacy: true,
    locale,
    messages: {
      [DEFAULT_LOCALE]: {
        test: "testValue",
      },
    },
  })

// Apply to vue app
app.use(i18n)

// Inside any component
<script setup>
const { locale } = useI18n();

console.log(locale); // Will print "en-US" 
<script/>
```

### System Info

```shell
System:
    OS: macOS 13.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 197.09 MB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.16.0 - ~/Library/Caches/fnm_multishells/89398_1674218098692/bin/node
    npm: 8.11.0 - ~/Library/Caches/fnm_multishells/89398_1674218098692/bin/npm
  Browsers:
    Brave Browser: 101.1.38.109
    Chrome: 109.0.5414.87
    Firefox: 107.0.1
    Safari: 16.2
  npmPackages:
    @intlify/vite-plugin-vue-i18n: 5.0.1 => 5.0.1 
    @vitejs/plugin-vue: 3.1.2 => 3.1.2 
    @vue/apollo-components: 4.0.0-alpha.20 => 4.0.0-alpha.20 
    @vue/apollo-composable: 4.0.0-beta.1 => 4.0.0-beta.1 
    @vue/apollo-option: 4.0.0-alpha.20 => 4.0.0-alpha.20 
    @vue/server-renderer: 3.2.37 => 3.2.37 
    vite: 3.2.2 => 3.2.2 
    vite-plugin-ssr: 0.4.54 => 0.4.54 
    vitest: 0.24.3 => 0.24.3 
    vue: 3.2.37 => 3.2.37 
[...]

PR #1316, opened 2023-01-20
fix: global locale is not respected by useI18n

Csszabi98:
Given the following config:
- `allowComposition: true`
- `legacy: true`

Calling `useI18n()` always results in
translations using the default language
"en-US".

Fixes: #1315


kazupon (approved):

That's good work!
Thanks!

PR #1366, opened 2023-04-04
Update breaking.md

Sojaner:
The `new` keyword is a mistake in calling of the `createI18n` function.


kazupon (approved):
Good catch!
Thanks!

PR #1374, opened 2023-04-17
Improve documentation about TypeScript Support

bchazelle-ubitransport:
Add some details about typing properly `createI18n()` with type hints for `Legacy` and `Options` type parameters


kazupon (approved):
Thank you so much! ❤️

PR #1375, opened 2023-04-18
docs: update vue-demi version

blackwindforce:


kazupon (approved):
Good catch!
Thanks! ❤️

Issue #1384, opened 2023-04-27
Type check with typescript v5 throws error in template $t


freakzlike:
### Reporting a bug?

After upgrade to typescript v5. The type check with `vue-tsc` throws errors with the usage of `$t` in vue template syntax. I think this might come from the `exports` definition in `package.json`, where you can set `types` per entry level. https://www.typescriptlang.org/docs/handbook/esm-node.html#packagejson-exports-imports-and-self-referencing

### Expected behavior

Don't throw a type error on check

### Reproduction

https://stackblitz.com/edit/vitest-dev-vitest-ztvbls?file=package.json

Run `npm run type-check-v5` to check types with `typescript@v5`

This should print the following errors:
```
src/Hello.vue:7:8 - error TS2339: Property '$t' does not exist on type '{ $: ComponentInternalInstance; $data: {}; $props: Partial<{}> & Omit<Readonly<ExtractPropTypes<{}>> & VNodeProps & AllowedComponentProps & ComponentCustomProps, never>; ... 10 more ...; $watch<T extends string | ((...args: any) => any)>(source: T, cb: T extends (...args: any) => infer R ? (args_0: R, args_1: R) => ...'.

7     {{ $t('test') }}
         ~~

src/Hello.vue:8:8 - error TS2339: Property '$t' does not exist on type '{ $: ComponentInternalInstance; $data: {}; $props: Partial<{}> & Omit<Readonly<ExtractPropTypes<{}>> & VNodeProps & AllowedComponentProps & ComponentCustomProps, never>; ... 10 more ...; $watch<T extends string | ((...args: any) => any)>(source: T, cb: T extends (...args: any) => infer R ? (args_0: R, args_1: R) => ...'.

8     {{ $t('test_with_label', { label }) }}
         ~~


Found 2 errors in the same file, starting at: src/Hello.vue:7
```

Run `npm run type-check-v4` to check types with `typescript@v4`

### System Info

```shell
System:
    OS: Linux 5.15 Ubuntu 20.04.6 LTS (Focal Fossa)
    CPU: (5) x64 AMD Ryzen 5 4500U with Radeon Graphics
    Memory: 3.42 GB / 6.78 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 18.15.0 - ~/.nvm/versions/node/v18.15.0/bin/node
    npm: 9.5.0 - ~/.nvm/versions/node/v18.15.0/bin/npm
[...]

PR #1388, opened 2023-05-01
fix(#1327): Fixed package.json exports to work with moduleResolution "bundle"

### LINKED ISSUE #1327: Typescript Import issue with NodeNext
### Reporting a bug?

When setting `"moduleResolution": "NodeNext",`  in `tsconfig.json` and trying to use this package will result in a error in VSCode `Could not find a declaration file for module 'vue-i18n'`. However everything runs fine, I think this has to do with how exports are being interpreted 

Changing the `package.json` exports to look like this seems to resolve the issue. Is this a change that would be acceptable?
  "exports": {
    ".": {
      "require": "./index.js",
      "import": "./dist/vue-i18n.esm-bundler.js",
      "types": "./dist/vue-i18n.d.ts"
    }
  },

### Expected behavior

Package should not show dev time TS errors when using `NodeNext` moduleResolution

### Reproduction

Quickest reproduction is to just create a TS vue app using Vite, and add `vue-i8n` to the project and update the `tsconfig.json` with `"moduleResolution": "NodeNext",`

### System Info

```shell
System:
    OS: Linux 5.15 Ubuntu 20.04.4 LTS (Focal Fossa)
    CPU: (16) x64 Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz
    Memory: 25.84 GB / 31.31 GB
    Container: Yes
    Shell: 3.2.2 - /usr/bin/fish
  Binaries:
    Node: 16.17.0 - ~/.nvm/versions/node/v16.17.0/bin/node
    Yarn: 1.22.19 - /usr/bin/yarn
    npm: 9.1.3 - ~/.nvm/versions/node/v16.17.0/bin/npm
  Browsers:
    Chrome: 102.0.5005.61
  npmPackages:
    @intlify/unplugin-vue-i18n: ^0.8.1 => 0.8.1
    @vitejs/plugin-vue: ^4.0.0 => 4.0.0
    vite: ^4.1.1 => 4.1.1
    vue: ^3.2.47 => 3.2.47
    vue-i18n: ^9.2.2 => 9.2.2
    vue-router: ^4.1.6 => 4.1.6
    vue-tsc: ^1.0.24 => 1.0.24
```

### Screenshot

_No response_

### Additional context

_No response_

### Validations

- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussions](https://github.com/intlify/vue-i18n-next/discussions)

lupas:
Fixes #1327 as proposed by [RebeccaStevens](https://github.com/RebeccaStevens) [here](https://github.com/intlify/vue-i18n-next/issues/1327#issuecomment-1484658084) as well as #1384.

With Typescript 5 & [vuejs/tsconfig 0.3.2](https://github.com/vuejs/tsconfig), `moduleResolution: 'bundle'` has become standard.

However, this package does not properly define typings in package.json which results in type-checking to fail & the following warning:
```
TS7016: Could not find a declaration file for module 'vue-i18n'. '/.../node_modules/vue-i18n/dist/vue-i18n.esm-bundler.js' implicitly has an 'any' type.
  There are types at '/.../node_modules/vue-i18n/dist/vue-i18n.d.ts', but this result could not be resolved when respecting package.json "exports". The 'vue-i18n' library may need to update its package.json or typings.
```

I'm not completely sure if the `node` and `browser` exports still serve any value, might wanna include them again if you see a reason for them.

### DIFF
--- a/packages/vue-i18n/package.json
+++ b/packages/vue-i18n/package.json
@@ -63,23 +63,18 @@
   },
   "exports": {
     ".": {
-      "types": "./dist/vue-i18n.d.ts",
-      "import": "./dist/vue-i18n.mjs",
-      "browser": "./dist/vue-i18n.esm-browser.js",
-      "node": {
-        "import": {
-          "production": "./dist/vue-i18n.node.mjs",
-          "development": "./dist/vue-i18n.mjs",
-          "default": "./dist/vue-i18n.mjs"
-        },
-        "require": {
-          "production": "./dist/vue-i18n.prod.cjs",
-          "development": "./dist/vue-i18n.cjs",
-          "default": "./index.js"
-        }
+      "import": {
+        "types": "./dist/vue-i18n.d.ts",
+        "node": "./index.mjs",
+        "default": "./dist/vue-i18n.esm-bundler.js"
+      },
+      "require": {
+        "types": "./dist/vue-i18n.d.cts",
+        "default": "./index.js"
       }
     },
     "./dist/*": "./dist/*",
+    "./index.mjs": "./index.mjs",
     "./package.json": "./package.json"
   },
   "funding": "https://github.com/sponsors/kazupon",


sodatea (commented):
This change is not necessary (and incorrect, because there's no `.cts` file in `dist`).

The current package.json already works with `"moduleResolution": "bundler"`.
It's just not yet tagged as `latest`. You can try it with `npm add vue-i18n@next` or `npm add vue-i18n@beta`

kazupon:
@sodatea 
Thanks for your follow up!

@lupas 
This PR is not necessary as `npm i vue-i18n@next` already supports it.
I would sorry to have spent your time.

If you have avoided #1327 and #1384 related issues around typescript, I would recommend installing the `next` tag version, (i.e. vue-i18n v9.3).
Currently in beta, this version has not changed much in terms of functionality from v9.2. Actually, It's used in nuxt i18n v8 beta. (I'm working on nuxt i18n right now)

PR #1399, opened 2023-05-09
fix: drop node v14

NozomuIkuta:
No related issues.

Node versions specified in the codebase and docs are inconsistent.
This PR is to unify Node versions to 16 for 2 reasons:

- only Node 16 is tested in CI (i.e. `.github/workflows/ci.yml`)
- the associated commit to `.github/workflows/ci.yml` is newer than other places (e.g. commit to `package.json`)

⚠️ This is my very first PR for this repository, so I might be missing something.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -22,10 +22,12 @@ It is of course fine to use non-English language, when you open a PR to translat
 - Work in the `src` folder and **DO NOT** checkin `dist` in the commits.
 
 - If adding new feature:
+
   - Add accompanying test case.
   - Provide convincing reason to add this feature. Ideally you should open a suggestion issue first and have it greenlighted before working on it.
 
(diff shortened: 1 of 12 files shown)


kazupon:

node v14 has already passed LTS, so I've changed the title name of this PR :)
https://github.com/nodejs/release#release-schedule

stanleyxu2005:
Hello contributors. 

I'd like to know, if the new version 9.3 or higher relies on any Node 16+ specific features? 
As you stated above 
> only Node 16 is tested in CI (i.e. .github/workflows/ci.yml)
It's fine to use Node 16 in CI, but it would be harmless for users that still running a Node 14 with their own risk?

I'm embarrassing to say but I'm still not able to upgrade all my old projects to Node 16+ due to breaking changes of Node, e.g. `require('fs')` must be `require('node:fs')`

PR #1426, opened 2023-06-10
doc: install latest version

wbxl2000:
Install the latest version would be better？


kazupon:

>Install the latest version would be better？

Basically we would prefer to install the latest version of ours. The documentation is also written assuming the latest version.

I'm glad for this PR of yours, but the documentation is written to install the latest version. So, I will close.

Thank you!

PR #1469, opened 2023-07-19
Add example code for deployed app for nuxt3-app-vue-i18n

lyqht:
I created a simple project following this integration guide and added [a custom github action that could translate locale jsons using deepL](https://github.com/lyqht/deepl-translate-github-action/). 

I thought that this will be a good example code project to be added into the integration guide.


kazupon (approved):
Thanks!

PR #1490, opened 2023-08-10
docs: diff with syntax highlighting in nuxt integration

jd-solanki:
Before:
image

After:
image


kazupon (approved):
Cool!
I’ve not known that notation so far.
Thanks!

PR #1499, opened 2023-08-21
docs: add vite-plugin-vue-i18n support vite version to warning notice

regchiu:
Avoid reader using wrong plugin when compiled and minified for production on vite 4.


kazupon (approved):
Thank you! ❤️

PR #1553, opened 2023-09-14
Update nuxt.config.ts in Nuxt 3 Minimal Starter Example

ram-you:
Add build.transpile to Nuxt config.
Solution  to address the problem documented in the GitHub issue at https://github.com/nuxt/nuxt/issues/23180.


kazupon (approved):
Thank you so much!

PR #1554, opened 2023-09-14
Handle falsy values in i18n.te()

BuuBux:
This pull request introduces a small fix to i18n-next that addresses an issue where

`i18n.te(null)` or `i18n.te(undefined)` will not provide false


kazupon:
Thanks!

Issue #1574, opened 2023-09-29
Typescript error when using pluralization and formatting


jangxx:
### Reporting a bug?

I'm trying to use pluralization and formatting together. This works as it should, but the TypeScript types seem to be incorrect, which makes it impossible to build the project properly.

### Expected behavior

The types match the actual call signature of the `t` function and I can successfully build my project.

### Reproduction

```ts
// messages.json
{
  "contacts": "{count} contact | {count} contacts"
}

// somewhere else
console.log(t("contacts", 1, { count: 1})) // prints "1 contact" correctly, but raises and error while building
```

The error in question:

```
  The last overload gave the following error.
    Argument of type 'number' is not assignable to parameter of type 'Record<string, unknown>'.

1   console.log(t("contacts", 1, { count: 1 }));
                              ~

  node_modules/vue-i18n/dist/vue-i18n.d.ts:1244:5
    1244     <Key extends string>(key: Key | ResourceKeys | number, named: NamedValue, defaultMsg: string): string;
             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    The last overload is declared here.
```

### System Info

```shell
System:
    OS: Linux 5.15 Linux Mint 21.1 (Vera)
    CPU: (32) x64 AMD Ryzen 9 7950X3D 16-Core Processor
    Memory: 45.44 GB / 61.96 GB
    Container: Yes
    Shell: 5.8.1 - /usr/bin/zsh
  Binaries:
    Node: 18.17.1 - /usr/bin/node
    npm: 9.6.7 - /usr/bin/npm
  Browsers:
    Chrome: 117.0.5938.92
  npmPackages:
    @vitejs/plugin-vue: ^4.0.0 => 4.1.0 
    @vue/tsconfig: ^0.1.3 => 0.1.3 
    vite: ^4.1.4 => 4.3.9 
    vue: ^3.2.47 => 3.2.47 
    vue-i18n: ^9.2.2 => 9.2.2 
    vue-router: ^4.1.6 => 4.1.6 
    vue-tsc: ^1.2.0 => 1.2.0
```

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/intlify/vue-i18n-next/blob/master/.github/CONTRIBUTING.md)
[...]

PR #1620, opened 2023-10-29
Create error.ts

Youngemmy5956:
Typescript error when using pluralization and formatting #1574

### DIFF
--- a/error.ts
+++ b/error.ts
@@ -1,0 +1,79 @@
+// Typescript error when using pluralization and formatting #1574
+
+
+
+
+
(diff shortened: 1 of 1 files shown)


kazupon (changes requested):
We should fix the below type definition

https://github.com/intlify/vue-i18n-next/blob/c37f6f03b72ba7ff3c340c058bd5037757013f73/packages/vue-i18n-core/src/composer.ts#L641-L976

Youngemmy5956:
will check it out

Issue #1630, opened 2023-11-05
Inconsistent Vue plugin `install` function interface between ts typing and the actual code ❗ 


k-paxian:
### Reporting a bug?

The first concerning issue 1️⃣ 

The `I18n` ts interface declares `install` function [here](https://github.com/intlify/vue-i18n-next/blob/d4909bfaf0e4b3dc984e06278c32d19dc9f9aa6c/packages/vue-i18n-core/src/i18n.ts#L254) as 
```ts
install(app: App, ...options: unknown[]): void
```
which is totally fine and aligned with the VueJs3 unit test [here](https://github.com/vuejs/core/blob/32bdc5d1900ceb8df1e8ee33ea65af7b4da61051/packages/runtime-core/__tests__/apiCreateApp.spec.ts#L295C3-L295C3) and aligned with the plugin installation contract - the `install` function expected to be synchronous ❗ 

But, the actual implementation function [here](https://github.com/intlify/vue-i18n-next/blob/d4909bfaf0e4b3dc984e06278c32d19dc9f9aa6c/packages/vue-i18n-core/src/i18n.ts#L589) is `async`
```ts
      // install plugin
      async install(app: App, ...options: unknown[]): Promise<void> {
```

The second concerning issue 2️⃣ 

When we assume the **first** issue is totally fine, since no one is concerned about it so far, because VueJs is actually calling
that `install` method **synchronously**.

When we try to call the async installation method it returns Promise which is never ending, always "pending"

```ts
import { createI18n } from 'vue-i18n'

const pluginInstance = createI18n({})
// this hangs the script execution, since Promise never will be resolved, it just waits indefinitely  🤷 
const installationPromiseResult =  await pluginInstance.install(app, {}) 
```

### Expected behavior

If you insists that the `install` method should be "async" for some reason, please do the right thing and return something / auto-resolve the returned Promise at least 😄 

### Reproduction

```ts
import { createI18n } from 'vue-i18n'

const pluginInstance = createI18n({})
// this hangs the script execution, since Promise never will be resolved, it just waits indefinitely 🤷 
const installationPromiseResult =  await pluginInstance.install(app, {}) 
```

### System Info

```shell
doesn't matter
```

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/intlify/vue-i18n-next/blob/master/.github/CONTRIBUTING.md)
- [X] Read the [Documentation](https://vue-i18n.intlify.dev/)
[...]

PR #1631, opened 2023-11-05
chore(fix): #1630 make the install function synchronous to be aligned…

k-paxian:
… with the Vue Plugin interface


kazupon (approved):
Sorry, late my reply. 🙇 
You're right!
Thanks!

PR #1649, opened 2023-11-30
feat!: type-safe messages

GalacticHypernova:
This PR adds type safe message translation when used with helpers such as t().

The reason it's a breaking change is because it's more strict, which could bring some edge case bugs with non-existent messages.

It is a draft for now because there is much work I need to do to get it to work.

#### Benefits:
This allows for much safer and convenient locale message translation by providing autocomplete suggestions to make sure no typos are present. Essentially a very useful DX update.

### DIFF
--- a/packages/vue-i18n-core/src/composer.ts
+++ b/packages/vue-i18n-core/src/composer.ts
@@ -118,8 +118,10 @@ import { isLegacyVueI18n } from './utils'
 
 export { DEFAULT_LOCALE } from '@intlify/core-base'
 
+type LocaleMessageKey = string
+
 // extend VNode interface
-export const DEVTOOLS_META = '__INTLIFY_META__'
+export const DEVTOOLS_META:LocaleMessageKey = '__INTLIFY_META__'
 
 /** @VueI18nComposition */
 export type VueMessageType = string | ResourceNode | VNode


kazupon:
Thank you for your contrivution!

First we would to confirm, I don't know from this PR what is the issue.
Could you first explanation on the details of the issue raised? 🙏

GalacticHypernova:
@kazupon sure thing!
Basically, the current translation functions like `t` don't have type safety, because they accept a simple string or a number. This would allow autocomplete suggestions for those functions. It's not really an issue, it's more so a DX improvement

kazupon:
@GalacticHypernova 
Thank you for your explanation!

>should this change VueMessageType or should this be added as an overload of t? Curious to hear your opinion on how to proceed with this. Both options have their advantages, and I would imagine backwards compatibility is preferred.

I am glad of your PR because I wanted to achieve fully  type-safe message!
Please, let's go ahead! :)

>I do not know how to generate unit tests for it, I'm still looking into it.

You can add type test to `test-dts` directory.

GalacticHypernova:
> You can add type test to `test-dts` directory.

I mean, this would require unit testing of type **inference**, not plain types. It partially relies directly on the typescript engine. Does vitest have something to deal with that?

kazupon:
> I mean, this would require unit testing of type inference, rather than regular types. It partially relies directly on the typescript engine. Does vitest have something to deal with that?

Yeah, vitest has type testing.
You can read it at here
https://vitest.dev/guide/testing-types.html

vue-i18n has its own type check utils for type testing at `test-dts`, but you can replace it with vitest since it uses it.

> Could you please tell me where the messages themselves are stored (as in, the messages object from which you're reading the translations)? I'm seeing multiple potential places, but thought I'd ask you. Is it the C in the ComposerTranslation? Or is it the readonly messages in the composer?

The messages themselves are stored in Compose's readonly messages, which is typed with the `Messages` passed in the Composer's generics.
ComposerTranslation is also typed with Composer's generics. ComposerTranslation is also typed with Composer's generics type, so that the `t` function can be used to complete the key.

GalacticHypernova:
Got it!
That means I'd probably be able to finish this PR's core functionality today, and then try the tests

GalacticHypernova:
@sagsagg I am planning to make it JSON compatible, yes. I'd have to see if json files get treated differently in terms of message parsing and adjust accordingly, but it is planned.

GalacticHypernova:
@kazupon does this make sense? From the tests that are already present the functionality appears to not really change, still getting all the right messages. Would appreciate a review from you so that in case something went wrong I won't need to rewrite everything. If I get the green light from you I could start changing all the translation functions.

kazupon:
@GalacticHypernova 
Sorry my late reply

> I realized the TranslationComposer has a Locales property. When changing it to Messages[Locales] the autocomplete works but TS errors saying Locales can't be used to index Messages. Can we allow it to error or should Locale be fixed? Perhaps we can extend ComposerTranslation from Composer or ComposerOptions? Although I don't quite know how that would affect messages from files.. This still needs a bit of experimentation.

Thank you for your contribution.

The value of `Locales` is passed via `Composer` and `ComposerOptions`. So `can't be used to index Messages` should not ignore the error because it will not be able to support the case where the type parameter is passed by that case. The type definition must be correctly supported so that all cases can be supported.

If `can't be used to index Messages` is causing wiht your implementation, you should add a type test to test-dts.
Adding it will avoid regression.

GalacticHypernova:
@kazupon 
Alright, I'll add a type test :+1: 
There doesn't seem to be a regression from my current testing, but perhaps it's worth investigating why Locale is seen as an empty string. And if you have any idea for a different implementation of it I'd be happy to hear.

GalacticHypernova:
@kazupon 
I've been looking into this a bit more, and I may have an idea that won't generate a type error, although I would like to discuss it with you regarding any possible performance impact it might bring, since i didn't get to test it enough like I did with my current idea. What is the type of ResourceKeys? If that's the thing that provides all the autocomplete options in the tests, it might be possible to "flatten" them into messages like I currently am trying with the actual Messages type.. but then again, I don't know how it will behave performance wise, so I'm gonna have to test it very carefully.

Tzyito:
Is this pr still going on? Look forward to

hongzzz:
Looking forward to this feature 😄

yarataliba:

GalacticHypernova:
@kazupon Was thinking about this a little more. Obviously reading the messages from the files with a typed solution won't be possible, considering types can't access runtime objects. How about generating a copy object, like during postinstall when nuxt generates all the types? The files could be taken and provided the given types which could be used for $t (and the other translation methods).

PR #1661, opened 2023-12-25
Register vue jsx types in tsx test

yyx990803:
In Vue 3.4, implicit global JSX type registration will be removed. Explicitly importing `vue/jsx` preserves the old behavior.


kazupon (approved):
Thank you for your feedback!|

> In Vue 3.4, implicit global JSX type registration will be removed. Explicitly importing vue/jsx preserves the old behavior.

I got it!

PR #1671, opened 2023-12-31
chore: improve Vue type definition appending

NozomuIkuta:
This PR is created based on [a conversation on Twitter](https://x.com/NozomuIkuta/status/1741107199001383376?s=20).

With this PR, we do no longer need to manage from which line of `<target>/src/vue.d.ts` should be appended to `<target>/dist/<target>.d.ts`. All the content below the marker string will be automatically appended.

Diffs in `vue.d.ts` files are mostly generated by Prettier.
All I did to those was to add "// --- THE CONTENT BELOW THIS LINE WILL BE APPENDED TO DTS FILE IN DIST DIRECTORY --- //" as the marker string.

---

This PR change build process, so please review carefully.


kazupon (approved):
Thank you so much!
Your PR is first contribution on this year :)

LGTM!

PR #1675, opened 2024-01-04
fix: memory leak

s00d:
Added support for arrays; this is needed if there's an object inside an array to solve the memory leak issue

https://github.com/nuxt-modules/i18n/issues/2629


kazupon:
Thank you for your contribution.
sorry my late reply. 🙇 

I appriciate your PR.
I believe memory leak issue will be resolved with #1677

I'll close your PR.
Thanks!

PR #1718, opened 2024-01-29
fix: key-value style messages broken after merging (#1717)

### LINKED ISSUE #1717: mergeLocaleMessage in petite 18n broken after 9.6.0 - converts to nested object by itself
### Reporting a bug?

After upgrading [petite-vue-i18n](https://github.com/intlify/vue-i18n-next/tree/master/packages/petite-vue-i18n) from 9.5.0 translations on app broke (messages not found). 

I'm using simple key, not nested objects (by default in petite). 

Screenshot from 2024-01-29 19-40-26

Translations are lazy loaded from .json files and merged.

```
const messages = await import(`./locales/${category}/${locale}.json`)
i18n.global.mergeLocaleMessage(locale, messages.default)
```

I debugged it via `i18n.global.getLocaleMessage('pl')` and noticed that my translations are nested, WTF :D

Screenshot from 2024-01-29 19-36-28

By downgrading version after version I discovered that issue is with 9.6.0, and noticed noticed this commit in changelog:
https://github.com/intlify/vue-i18n-next/pull/1601/commits/edaf49806d8da69d350d12ce3527c0031c15875b

Using `flatConfig` in i18n config does not help, and I guess should not be used in petite anyway.

I've commented this line in petite-vue-i18n dist source:
```
handleFlatJson(_message[key]);
```
and it works.

Then I noticed that in `setLocaleMessage` there is condition
```
if (flatJson) {
```
for that `handleFlatJson`, but it's missing in mergeLocaleMessage. **Why?** After adding this, again everything works. Was this on purpose? I case this is a typo, I will provide pull request. If not - @kazupon please provide other solution :( 

### Expected behavior

Keys like 'a.b.c' should be treated like strings, not transformed into object.

### Reproduction

Described above.

### System Info

```shell
Described above.
```

### Screenshot

_No response_

### Additional context

_No response_

### Validations

- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussions](https://github.com/intlify/vue-i18n-next/discussions)

chojnicki:
resolves #1717

more info in issue

### DIFF
--- a/packages/vue-i18n-core/src/composer.ts
+++ b/packages/vue-i18n-core/src/composer.ts
@@ -2435,9 +2435,11 @@ export function createComposer(options: any = {}, VueI18nLegacy?: any): any {
       __legacy && __legacy.mergeLocaleMessage(locale, message)
     }
     const _message = { [locale]: message }
-    for (const key in _message) {
-      if (hasOwn(_message, key)) {
-        handleFlatJson(_message[key])
+    if (flatJson) {
+      for (const key in _message) {
+        if (hasOwn(_message, key)) {
+          handleFlatJson(_message[key])
+        }
       }
     }
     message = _message[locale]
--- a/packages/vue-i18n-core/test/issues.test.ts
+++ b/packages/vue-i18n-core/test/issues.test.ts
@@ -1273,3 +1273,19 @@ test('issue #1615', async () => {
     `false (...but this should be false)`
   )
 })
+
+test('issue #1717', async () => {
+  const en = {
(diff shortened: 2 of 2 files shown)


kazupon (changes requested):
Can you add the test for your issue case, please? 🙏 

You should add as a reproducible case test case for this issue.
With that test code we can recognize the issue. It will also prevent regressions.

chojnicki:
here you go @kazupon

kazupon (approved):
Good contritubtion!
Thanks!

PR #1732, opened 2024-02-09
chore(docs): change eslint plugin's URL to the right github repo

twolfvb:
The current one gives a 404 from github pages.


kazupon (approved):
Good catch! 
Thanks!

PR #1848, opened 2024-05-10
docs: fix typo on vue i18n name

r-moret:
Just a small fix in the _Getting Started_ section of the docs where it said "Vue Router" but I think it meant "Vue I18n" 🙂


kazupon (approved):
good catch!
Thanks!

PR #1866, opened 2024-05-31
Update tools.md

felixhaeberle:
Hey @kazupon,
hope you are doing good☀️

We recently renamed our VS Code extension to "Sherlock" and made the ecosystem experience even better.
Let me know what needs to be changed in order to merge this.

Have a nice day!:)
Greetings from 🇩🇪

### DIFF
--- a/docs/ecosystem/tools.md
+++ b/docs/ecosystem/tools.md
@@ -1,10 +1,10 @@
 # Third-party tooling
 
-## inlang
+## Sherlock – i18n inspector (VS Code extension)
 
-inlang is the ecosystem to adapt and expand software products to new markets, and that has [VSCode extensions](https://marketplace.visualstudio.com/items?itemName=inlang.vs-code-extension).
+Sherlock is a VS Code extension that helps you to extract, edit & inspect your i18n keys in your codebase. It is available in the [VS Code Marketplace](https://marketplace.visualstudio.com/items?itemName=inlang.sherlock).
 
-inlang give awesome developer experience for your i18n development.
+Sherlock is part of the inlang ecosystem, which gives you multiple apps with awesome developer experience for your i18n development, like a [Figma plugin](https://inlang.com/m/gkrpgoir/app-parrot-figmaPlugin), a [web editor](https://inlang.com/m/tdozzpar/app-inlang-finkLocalizationEditor) or a [GitHub Action](https://inlang.com/m/3gk8n4n4/app-inlang-ninjaI18nAction).
 
 
 ## i18n Ally
--- a/docs/ecosystem/tools.md
+++ b/docs/ecosystem/tools.md
@@ -2,7 +2,7 @@
 
 ## Sherlock – i18n inspector (VS Code extension)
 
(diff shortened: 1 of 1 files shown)


docs/ecosystem/tools.md
```diff
@@ -1,10 +1,10 @@
 # Third-party tooling
 
-## inlang
+## Sherlock – i18n inspector (VS Code extension)
 
-inlang is the ecosystem to adapt and expand software products to new markets, and that has [VSCode extensions](https://marketplace.visualstudio.com/items?itemName=inlang.vs-code-extension).
+Sherlock is a VS Code extension that helps you to extract, edit & inspect your i18n keys in your codebase. It is available in the [VS Code Marketplace](https://marketplace.visualstudio.com/items?itemName=inlang.sherlock).
```
kazupon, line 5: `https://marketplace.visualstudio.com/items?itemName=inlang.sherlock` is dead link
Where would it?

kazupon:
And, I noticed that there was no mention of vue-i18n in inlang vue docs.
https://inlang.com/c/vue

In the react docs, there are plugins for i18next and next-intl.
https://inlang.com/c/react

`3rd Party Tooling` is the section for use vue-i18n or put tools that work with vue-i18n.
Why is there no vue-i18n in your docs or not use vue-i18n ?

kazupon:
sorry, I also missed about having this page. There it mentions vue-i18n.
https://inlang.com/m/r7kp499g/app-inlang-ideExtension/supported-i18n-libraries

However, I would like to see a small mention of vue-i18n on the Vue page.
I also think inlang is a great tool. I would like to be part of the ecosystem, working on the same i18n issues.

kazupon:
And, If possible, we would also like hopely to put about nuxt i18n, which integrates vue-i18n.

felixhaeberle:
I talked to the team and we removed the entry in order to achieve first-class support through a dedicated plugin we are building both for `vue-i18n` & `nuxt-i18n`. So this was probably a bit too early for us to remove the listing but will have a comeback with dedicated first-class [plugin](https://inlang.com/documentation/plugin/guide).

Progress can be tracked here:
- https://github.com/opral/inlang-sherlock/issues/70
- https://github.com/opral/inlang-sherlock/issues/71

Btw, this is not only Sherlock (VS Code extension) related. With a plugin the library/content gets enabled to be [part of every inlang app](https://inlang.com/c/apps) in the inlang ecosystem.

kazupon:
@felixhaeberle 
Thank you for considering putting vue-i18n and nuxt-i18n in inlang.
We hope the issue will be realized.

> With a plugin the library/content gets enabled to be [part of every inlang app](https://inlang.com/c/apps) in the inlang ecosystem.

I wonder what that means. Does it mean that vue-i18n and other i18n libraries will be imported as inlang apps and disappear like this issue situation? 🤔

felixhaeberle:
> I wonder what that means.

An inlang plugin is a connector to the ressource files & markup used a vue-i18n / nuxt-i18n project. vue-i18n & nuxt-i18n users can then profit from all inlang apps because inlang apps "understand" the code & ressource files used + can then also modify them through apps (like Sherlock).

kazupon:
Thank you for your explanation!
Okay, I belive that your efforts will return for those involved in i18n libraries and toolings :)

Issue #882, opened 2022-01-16
Custom pluralization rule is not working


RollingTL:
### Reporting a bug?

Hi,
I am trying to use custom pluralization function but it seems it is not working as expected:

The result is:
```
<p>1 машина</p>
<p>2 машины</p>
<p>4 машины</p>
<p>12 машины</p>
<p>21 машины</p>
```

### Expected behavior

```
<p>1 машина</p>
<p>2 машины</p>
<p>4 машины</p>
<p>12 машин</p>
<p>21 машина</p>
```

12 and 21 result is wrong.

### Reproduction

[https://github.com/RollingTL/Testing-intlify-vue-i18n-next-custom-pluralization](url)

### System Info

```shell
Mac OS
```

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/intlify/vue-i18n-next/blob/master/.github/CONTRIBUTING.md)
- [X] Read the [Documentation](https://vue-i18n.intlify.dev/)
- [X] Check that there isn't [already an issue](https://github.com/intlify/vue-i18n-next/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/intlify/vue-i18n-next/discussions)
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

RollingTL:
I played with pluralization function and I think that it never called.
And one question what is the purpose of second and third argument? ()choicesLength, orgRule)

RollingTL:
After reading sources I have found the problem 
Instead of **pluralizationRules**  as in documentation the option should be **pluralRules**.


```
const i18n = createI18n({
  locale: 'ru',
  // the custom rules here ...
  **pluralRules**: {
    ru: customRule
  },
  messages: {
    ru: {
      car: '0 машин | {n} машина | {n} машины | {n} машин',
      banana: 'нет бананов | {n} банан | {n} банана | {n} бананов'
    }
  }
})
```
It is necessary to change documentation.

kazupon:
Thank you for your feedback.

About custom pluralization, vue-i18n@v9.x or later have a breaking change 
You need to check docs here:
https://vue-i18n.intlify.dev/guide/migration/breaking.html#pluralization-api-return-value

In composition API, we don't mention custom pluralization.
We should improve the docs about custom pluralization.
[...]

PR #1873, opened 2024-06-13
Added a comment about pluralRules to the code pluralization docs

M1h4n1k:
I removed meaningless comment and replaced it with more useful comment about `pluralRules` for Composition api.
I am just looking for the code and I'm not really reading through the documentation if the code is understandable, and seems like many people do the same https://github.com/intlify/vue-i18n/issues/882. However, in the docs there is a very important note. So, there is a higher chance that people will notice it if it's put directly into the code. 

_I've actually done the same mistake in 2 different projects. Luckily, at the second time I remembered that there was something wrong with docs and it didn't take me very long, but at the first time I spent quite a lot of time_

### DIFF
--- a/docs/guide/essentials/pluralization.md
+++ b/docs/guide/essentials/pluralization.md
@@ -152,7 +152,7 @@ like the the following locale:
 ```js
 const i18n = createI18n({
   locale: 'ru',
-  // the custom rules here ...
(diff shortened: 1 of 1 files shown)


kazupon (approved):
Sorry, my late response
We appreciate your PR!
Thank you !

Issue #1215, opened 2022-11-05
`$t` does not have TypeScript auto-completions


innocenzi:
### Reporting a bug?

The `$t` global function in SFCs' templates don't seem to support global TypeScript definitions.

![CleanShot 2022-11-05 at 23 02 35](https://user-images.githubusercontent.com/16060559/200142987-a1115247-4dbf-4d44-ab54-10f7a7fdb732.png)

The [documentation](https://vue-i18n.intlify.dev/guide/advanced/typescript.html#global-resource-schema-type-definition) regarding this topic doesn't mention `$t` but its definition look like it should support auto-completion.

### Expected behavior

`$t` should provide auto-completion through the overriden `DefineLocaleMessage` interface, like `t` exported from `useI18n` does.

![CleanShot 2022-11-05 at 23 03 33](https://user-images.githubusercontent.com/16060559/200143037-bca4a350-ae87-4b89-94a5-bed5650ccb0b.png)

### Reproduction

Create a `i18n.json` file with the following content: 

<details>
<summary>Details</summary>

```json
{
	"en": {
		"login": {
			"welcome_text_title": "Welcome back!",
			"welcome_text_help_text": "Enter your credentials to access the app.",
			"email_field_placeholder": "Enter your email",
			"password_field_placeholder": "Enter your password",
			"sign_in_button_label": "Sign in",
			"quick_login_cta_text": "Wanna login quickly?",
			"quick_login_link_text": "Click here."
		}
	}
}

```

</details>

Create a `i18n.ts` file with the following content:

```ts
import { createI18n } from 'vue-i18n'
import messages from './i18n.json'

export type MessageSchema = typeof messages['en']

declare module 'vue-i18n' {
	export interface DefineLocaleMessage extends MessageSchema {}
}

const i18n = createI18n<[MessageSchema], 'en'>({
	locale: 'en',
	inheritLocale: true,
	messages,
})

export default i18n
```

Create a Vue 3 application, import `./i18n.ts` and register the exported plugin.

### System Info

```shell
System:
    OS: macOS 12.6
    CPU: (10) arm64 Apple M1 Pro
    Memory: 91.94 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.11.0 - /opt/homebrew/bin/node
    Yarn: 1.22.18 - /opt/homebrew/bin/yarn
    npm: 8.19.2 - /opt/homebrew/bin/npm
  Browsers:
[...]

PR #1883, opened 2024-07-04
fix(types): `$t` types

mitjans:
This PR addresses 2 problems with Typescript types:

- #1215: By removing the type `Path` from the `key: Key | ResourceKeys | Path`, we get type inference to work again and suggest valid translations when using globally injected `$t`. Since the type `Key extends string` already extends from `string`, it shouldn't cause any regression. For the type inference to work when using `$t`, you have to extend the module types:
	```ts
	import en from '@src/assets/locales/en.json';
	
	type Messages = typeof en;
	
	declare module 'vue-i18n' {
	  export interface DefineLocaleMessage extends Messages {}
	}
	```
- When fixing the first issue, I found that nested keys are treated as translations by themselves. For example:
	```json
	{
	  "this": {
	    "is": {
	      "a": {
	        "nested": {
	          "key": "This is a nested key"
	        }
	      }
	    }
	  }
	}
	```
	When having this `en.json` locale file, type inference will suggest a combination of all keys:
	image
	But this is not correct, only `this.is.a.nested.key` should be suggested. I've created two new types to solve this:

1. `JsonPaths`: Will return leaf keys only, and not suggest in-between keys from a JSON or Javascript object. I've called it Json because this should be used when extracting translations from JSON files.

2. `TranslationsPaths`: This will simply remove the first level of a JSON or JS object and return the `JsonPaths` from the second level onwards. This is used for translations coming directly from the composer, where the top-level key is the language.

I've tried not to modify existing types to avoid breaking something, and I just created these 2 new types. Here're two images of the new behavior both in `$t` and when using the composer:
image
image

### DIFF
--- a/packages/core-base/src/types/utils.ts
+++ b/packages/core-base/src/types/utils.ts
@@ -30,6 +30,14 @@ export type LocaleRecord<T extends any[], R> = {
 
 export type First<T extends readonly any[]> = T[0]
 
+export declare type JsonPaths<T, Key extends keyof T = keyof T> = Key extends string ? 
+  T[Key] extends Record<string, any> ?
+    `${Key}.${JsonPaths<T[Key]>}`
+  : `${Key}`
+: never;
+export declare type TranslationsPaths<T extends object, K extends keyof T = keyof T> = K extends string
+  ? JsonPaths<T[K]> : never;
+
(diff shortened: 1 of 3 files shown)


kazupon (approved):
Wow! Great works!

Your PR will improve i18n DX!

Thanks!

PR #1888, opened 2024-07-14
fix: vue-i18n type definition for vue package

kazupon:


BobbieGoede:
I think we will need a patch release for v9 to also change the type augmentations to use `'vue'` 😥 (relevant https://github.com/nuxt/nuxt/pull/28542)

Due to Vue Router changing its augmentation, Nuxt is doing so too, mixed augmentation will break projects with v9.

BobbieGoede:
@kazupon 
I've created a `v9` branch based on e37cba0657e9f5896203a07f062936362d72dc17 which includes this change as well as c07fbe37fb84a2fa9d4eaf20c8a3952c6bdbfa34 here https://github.com/BobbieGoede/vue-i18n-next/tree/v9

hinogi:
problem persists with v10.0.0 and vue 3.5

BobbieGoede:
@hinogi 
Can you open a new issue and provide a minimal reproduction? 🙏

hinogi:
I guess I am unable to do that. But I use Quasar and vite and I get `TS2339 Property $t does not exist on type 'CreateComponentPublicInstanceWithMixins<ToResolvedProps<__VLS_PublicProps, { "update:modelValue": (modelValue: string) => any }>...>`

Narretz:
Why was this changed? This breaks the legacy API.

Narretz:
Ok, so this happens when other packages augment `@vue/runtime-core` instead of `vue`. For me,this was my own app, vue-router, and vue/apollo-option.

Still, this change/info is missing from the changelogs, as far as I can see: https://github.com/intlify/vue-i18n/blob/v9/CHANGELOG.md and https://github.com/intlify/vue-i18n/blob/c580c3a5baa6cc3e60e962ff8f6e95ad2372b135/CHANGELOG.md

BobbieGoede:
@Narretz 
We (and all the other packages that did this) should never have augmented `@vue/runtime-core`, this incorrect augmentation only worked so long as all other packages did the same thing.

At some point packages were moving to the [correct augmentation, augmenting `vue`](https://vuejs.org/guide/typescript/options-api.html#augmenting-global-properties), which gradually broke projects that have mixed augmentation usage. This is only a breaking change if your project or dependencies were using the incorrect augmentation.

PR #1896, opened 2024-07-17
feat: supports selecting the overwrite order of merge messages

supercpq:
Hello, I added this code to provide users facing this situation with more options: whether to replace the existing message on the i18n instance with the passed message.
And this way of writing can keep the function calls in the vue-i18n npm package unaffected.
The background reason is that I have an npm package and a Vue3 code base. When app.use this npm package, I passed the i18n instance of the Vue3 code base to the install function of the npm package, and then called mergeLocaleMessage, which caused the message of the npm package to overwrite the value of the same key in the Vue3 code base, but I hope that the message in the Vue3 code base has a higher priority, so I thought of this way.
I hope it can be adopted, thank you.

### DIFF
--- a/packages/shared/src/messages.ts
+++ b/packages/shared/src/messages.ts
@@ -2,7 +2,7 @@ import { isArray, isObject } from './utils'
 
 const isNotObjectOrIsArray = (val: unknown) => !isObject(val) || isArray(val)
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-export function deepCopy(src: any, des: any): void {
+export function deepCopy(src: any, des: any, isRetainExistMessage?: boolean): void {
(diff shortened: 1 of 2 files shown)


supercpq:
That means I need to call mergeLocaleMessage twice, the first time passing in what I need to merge, and the second time passing in what I originally wanted, to achieve the overwriting situation. This is indeed a solution, but if it can be directly supported from mergeLocaleMessage, I believe it will be more readable and have higher performance.

supercpq:
Compared with the original version:
Original version: When the key value is the same, the incoming message will overwrite the original message on the i18n instance
New version: When the key value is the same, if `isRetainExistMessage=true` is passed in, the original message on the i18n instance will not be changed
In other cases, the two boards are the same 😃

PR #1925, opened 2024-08-28
Improved the global and local Scope descriptions

pankajrlal:
The earlier text seemed a little difficult to understand. Made some changes to make it comprehensible.


kazupon (approved):
Thank you so much!

PR #1971, opened 2024-09-23
Add petite-vue-i18n package to README

fdc-viktor-luft:
Additionally, fixed wrong link of `vue-i18n-bridge` package, and added trailing pipes for table.


kazupon (approved):
Good catch!
Thanks!

PR #1999, opened 2024-10-31
Change verbosity of Not found parent scope use global scope

Tofandel:
This warning has been annoying everyone for more than 5 years now. As a first step, let's switch it to a debug message at least
https://github.com/intlify/vue-i18n/discussions/851

### DIFF
--- a/packages/vue-i18n-core/src/i18n.ts
+++ b/packages/vue-i18n-core/src/i18n.ts
@@ -4,8 +4,7 @@ import {
   isBoolean,
   isEmptyObject,
   isPlainObject,
-  makeSymbol,
-  warn
+  makeSymbol
 } from '@intlify/shared'
 import {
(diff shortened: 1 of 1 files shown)


kazupon:
Thank you for your feedback!

As for the warning, this is intentionally output to promote understanding of which scope i18n-t is being used. This warning is only printed in development mode, not in production.
If you want to avoid this, you can use global scope or create your wrapper.

Thanks!

Tofandel:
But it should not be a warning, it should be an information or debug message.

Just check the discussion to understand that everyone is in favor of not having this be a warning because

1. It floods the console in development for no valid reason, and doesn't indicate **where** the problem is coming from
2. It is not a valid reason for a warning because i18n will still fallback to the global scope in all cases, which is absolutely expected behavior

PR #2017, opened 2024-11-15
breaking: drop `$tc` and `tc`

kazupon:
## breaking changes

Translate pluralization API `$tc` and `tc` were deprecated in v10.
The code that has interface and warning was left in place, we remove completely in v11, as it is no longer needed.


BobbieGoede (approved):
Looks good to me!

We should add a breaking changes page for v11 that mentions the full removal of the deprecated features, I forgot to mention it in the previous PR 😅

While checking for the occurrence of `$tc` and `tc` I noticed there's outdated Japanese documentation in `docs/.ja`, are these still being used some place? If not, we could clean it up in a separate change.

kazupon:
> We should add a breaking changes page for v11 that mentions the full removal of the deprecated features, I forgot to mention it in the previous PR 😅

Oops.
I forgot it. 😓 
I've just added docs
https://github.com/intlify/vue-i18n/pull/2017/commits/6c8c9a323e267d390e2775a5cc5d8e7fb46ea000

BobbieGoede (approved):

PR #2045, opened 2024-12-05
feat!: deprecate `v-t` custom directive

kazupon:
## Reason

The advantage of `v-t` was that it could optimize performance using the vue compiler transform and the pre-translation of `vue-i18n-extension`.

This feature was supported from Vue 2.
About details see the blog article: https://medium.com/@kazu_pon/performance-optimization-of-vue-i18n-83099eb45c2d

In Vue 3, due to the Composition API, the pre-translation of `vue-i18n-extension` is now limited only for global scope.

In addition, Vue 3 Virtual DOM optimization has been introduced, and the optimization provided by `vue-i18n-extension` is no longer very effective. We need to require settings for SSR, the benefits of using `v-t` have disappeared. And DX of templates using `v-t` is not good. Custom directives do not work with key completion in editors (e.g. vscode).

For compatibility, `v-t` mode still works in v11, but will be removed entirely in v12, so `v-t` will not work after that version.

### DIFF
--- a/docs/.vitepress/config.mts
+++ b/docs/.vitepress/config.mts
@@ -195,16 +195,16 @@ function sidebarGuide() {
       collapsible: true,
       items: [
         {
-          text: 'Breaking Changes in v9',
-          link: '/guide/migration/breaking'
+          text: 'Breaking Changes in v11',
+          link: '/guide/migration/breaking11'
         },
         {
           text: 'Breaking Changes in v10',
           link: '/guide/migration/breaking10'
         },
         {
-          text: 'Breaking Changes in v11',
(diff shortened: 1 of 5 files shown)


BobbieGoede (approved):
LGTM!

putianyi889:
> In addition, Vue 3 Virtual DOM optimization has been introduced, and the optimization provided by vue-i18n-extension is no longer very effective. We need to require settings for SSR, the benefits of using v-t have disappeared.

Can you elaborate? What is the recommended choice then, in terms of performance? Is there an in-depth discussion of this change?

PR #2058, opened 2024-12-27
feat: reactive `availableLocales` using computed property

### LINKED ISSUE #1893: Make availableLocales reactive
### Clear and concise description of the problem

Using availableLocales in a component is a list of strings that doesn't update reactively when adding new messages.

```js
const { availableLocales } = useI18n();  // string[]

i18n.global.setLocaleMessage('ca', messages); // Does not trigger an update to availableLocales 
```

### Suggested solution

Would be useful if availableLocales was a computed list so any list or selector or whatever that is using it can update automatically

```js
const { availableLocales } = useI18n(); // ComputedRef<string[]>
```

Not sure if changing line https://github.com/intlify/vue-i18n/blob/e9ab3e0933b8ac9cb859578235b363876065fdf2/packages/vue-i18n-core/src/composer.ts#L2559 to something like the following would do the trick.

```js
return computed(() => Object.keys(_messages.value).sort());
```

### Alternative

_No response_

### Additional context

_No response_

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/intlify/vue-i18n/blob/master/.github/CONTRIBUTING.md)
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

mnenie:
resolves #1893 

so, I had the same problem and as a solution I implemented the idea suggested by the man who created the issue using `computed` in `availableLocales` at `composer.ts`
also I wrote some tests and change some things in `legacy.ts` to accommodate this of course
now its works great :)

### DIFF
--- a/packages/vue-i18n-core/test/legacy.test.ts
+++ b/packages/vue-i18n-core/test/legacy.test.ts
@@ -66,6 +66,22 @@ describe('availableLocales', () => {
     i18n.setLocaleMessage('ru', { hello: 'Привет мир!' })
     expect(i18n.availableLocales).toEqual(['en', 'ja', 'fr', 'ru'].sort())
   })
+  test('trigger new value with watchEffect', async () => {
+    const i18n = createVueI18n({
+      messages: {
+        ja: {}
+      }
+    })
+    let locales = [] as string[]
+
+    watchEffect(() => {
+      locales = i18n.availableLocales
+    })
+
+    i18n.setLocaleMessage('en', { hello: 'Hello' })
+    await nextTick()
+    expect(locales.sort()).toEqual(['en', 'ja'].sort())
+  })
 })
 
 test('missing', () => {
--- a/packages/vue-i18n-core/test/legacy.test.ts
+++ b/packages/vue-i18n-core/test/legacy.test.ts
@@ -44,16 +44,28 @@ test('fallbackLocale', () => {
   expect(i18n.fallbackLocale).toEqual('ja')
 })
 
-test('availableLocales', () => {
-  const i18n = createVueI18n({
-    messages: {
-      en: {},
-      ja: {},
-      ru: {},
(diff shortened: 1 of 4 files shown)


packages/vue-i18n-core/test/legacy.test.ts
```diff
@@ -44,16 +44,28 @@ test('fallbackLocale', () => {
...
+    const i18n = createVueI18n({
+      messages: {
+        en: {},
+        ja: {},
+        fr: {}
+      }
+    })
+    expect(i18n.availableLocales).toEqual(['en', 'ja', 'fr'].sort())
+  })
+  test('reactive update', () => {
```
kazupon, line 58: It is necessary to test with `watch` / `watchEffect` to see if the reaction is taking place.

kazupon (changes requested):
Thank you for your contributing!
I've just reviewed your PR!
Please, let's check it out :)

mnenie:
> Thank you for your contributing! I've just reviewed your PR! Please, let's check it out :)

hi @kazupon !
so, I made it
I hope that's what you asked for 💯 :)

BobbieGoede:
Just checking, this would be a breaking change right?

mnenie:
> Just checking, this would be a breaking change right?

maybe yea
so, now we have a reactive scenario with `availableLocales ` and the type of usage will change to `.value` in `setup`

kazupon:
@BobbieGoede 
> Just checking, this would be a breaking change right?

Thank you for making me realize!
Yeah, This PR has breaking changes.
So, I've made `v11` branch.
https://github.com/intlify/vue-i18n/tree/v11

v11 branch will be included bug fixes and backward-compatible features until vue-i18n v12
`master` branch is v12 develop branch.

kazupon (approved):

PR #2065, opened 2025-01-03
perf: more optimize bundle size

kazupon:
With the introduction of Jit compilation in v9, code generation is no longer necessary for dist files, such as `@intlify/core`, `vue-i18n` and `petite-vue-i18n` globally. 

This PR is optimized to reduce bundle size for those dist files.


packages/core-base/src/format.ts
```diff
@@ -177,7 +177,7 @@ function resolveValue<Message = string>(
     node as Node,
     PROPS_VALUE
   ) as MessageType<Message>
-  if (resolved) {
+  if (resolved != null) {
```
kazupon, line 180: I found bug for Literal case, such as  `''`.
I've fixed it.

PR #2081, opened 2025-01-20
fix(installation.md): update installation commands for package manager

Vincere1st:
The commands are not updated with the new version.


kazupon (approved):
Good catch!
Thanks!

PR #2104, opened 2025-02-08
Fix: Synchronous Watcher Issue Caused by Locale Setting Order

lcjnil:
reproduction: https://codesandbox.io/p/devbox/vue-i18n-9-template-forked-ffq44z

This PR adjusts the order of operations in the locale setter to rectify the issue where, during locale updates, a synchronous watcher may trigger after modifying the _locale.value (as seen in [this file](https://github.com/intlify/vue-i18n/blob/897829bff49fe4694cc9fc0c571d2d42652e4b9f/packages/vue-i18n-core/src/composer.ts#L2113), but before updating the _context. This sequence led to errors because the translation mechanism relies exclusively on _context, not _locale.

### DIFF
--- a/packages/vue-i18n-core/src/composer.ts
+++ b/packages/vue-i18n-core/src/composer.ts
@@ -2110,8 +2110,8 @@ export function createComposer(options: any = {}): any {
   const locale = computed({
     get: () => _locale.value,
     set: val => {
+      _context.locale = val
       _locale.value = val
-      _context.locale = _locale.value
     }
   })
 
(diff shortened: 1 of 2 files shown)


kazupon:

The CI is not working correctly due to corepack issue.
I have pushed a commit to the `master` branch that fixes the issue.
If you pull in that commit, it should work correctly.
Would you please do that? 🙏

kazupon:
never mind for my comment.

vue-i18n GitHub PR has just been able to update from branches.

CI results are all green :) 
Thanks!

PR #2109, opened 2025-02-12
drop node v18

babu-ch:
close https://github.com/intlify/vue-i18n/issues/2099

hi
I saw a post on the chibivue server and came to help.
I confirmed that build and test are working.

Unclear points:
How to run tests in workflows
Do I also need to modify package.json under `packages`?


kazupon (approved):
Thank you so much!

Yeah, That's right!
Your contribution is so perfect! 💯 
I'll merge your PR!

Thanks!

PR #2112, opened 2025-02-15
ship ESM only

babu-ch:
closes https://github.com/intlify/vue-i18n/issues/2101

By the way, is `docs/.ja` not used?
If so, would you like me to remove it from this PR?
When I grep with cjs, it gets matched


docs/guide/extra/dist.md
kazupon, line 1: We want to remove for cjs, but vue-i18n docs have to support v9 - v12.
So, We need to mention for cjs dropping on v12 as important tip.

kazupon:

I have one more favor to ask 🙏 
We want to ship vue-i18n dist files as `.js`, not `.mjs`.
Let's try to work on it. 😉

kazupon:
And, We need to find out about code size increasing
https://github.com/intlify/vue-i18n/pull/2112#issuecomment-2661457233

kazupon:
> By the way, is docs/.ja not used?

Ah, sorry.
docs/.ja is not maintained, so you can remove another PR.

babu-ch:
@kazupon 
>We want to ship vue-i18n dist files as .js, not .mjs.

Fixed in commit 
https://github.com/intlify/vue-i18n/pull/2112/commits/5c725d40f2b882be2eb69ab3fa11c93e5251983e

babu-ch:
@kazupon 
> And, We need to find out about code size increasing

Maybe it was because I was working on an old branch
merged the latest

kazupon:
@babu-ch 
Well done!
Perfect! 💯 

Thank you so much!

PR #2118, opened 2025-02-18
docs: clarify `v-t` limitations and reactivity behavior

docnoe:
- Improved explanation of `v-t`'s limitations regarding inline HTML and dynamic template expressions.
- Clarified that `v-t` supports reactivity but is less flexible compared to `$t`.
- Enhanced wording for better readability and accuracy.

This improves the documentation to better reflect the behavior of `v-t` and avoid potential misunderstandings.


kazupon (approved):
Thank you!

PR #2128, opened 2025-03-09
fix: update MountOptions.components type

johnsoncodehk:
👋 Here is a type fix to make https://github.com/vuejs/core/pull/12935 pass ecosystem CI.


kazupon (approved):
@johnsoncodehk 
Thank you! 💚

PR #2167, opened 2025-04-18
fix: update slot key filtering to exclude internal keys

edison1105:
We will add another internal key `__`, see https://github.com/vuejs/core/pull/13215/files


kazupon (approved):
Thanks!

kazupon:
Ah, Sorry, I've not backed port from master branch to v11 branch.
I'll do it.

Issue #2172, opened 2025-04-27
back-port `n` (`$n`) part option for v11


kazupon:
The below comments are our notes:

- back-port for v11

_Originally posted by @kazupon in https://github.com/intlify/vue-i18n/issues/2170#issuecomment-2833370761_

PR #2175, opened 2025-04-28
Part options support $n - v11

mauryapari:
Back ported part option functionality for $n().

Would close https://github.com/intlify/vue-i18n/issues/2172

### DIFF
--- a/packages/vue-i18n-core/test/composer.test.ts
+++ b/packages/vue-i18n-core/test/composer.test.ts
@@ -5,6 +5,7 @@
 
 // utils
 import * as shared from '@intlify/shared'
+import { pluralRules as _pluralRules } from './helper'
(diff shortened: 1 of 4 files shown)


packages/vue-i18n-core/test/composer.test.ts
```diff
@@ -5,41 +5,41 @@
...
   const actual = await vi.importActual<object>('@intlify/shared')
   return {
     ...actual,
     warn: vi.fn()
   }
 })
-import { pluralRules as _pluralRules } from './helper'
 
 import {
+  compile,
```
mauryapari, line 18: Import statements were moved due to the file linting while committing new test cases

kazupon (approved):
Thank you!
LGTM!

Issue #2171, opened 2025-04-27
support the part options for `d` (`$d`)


kazupon:
- support the part options for `d` (`$d`) 
- back-port for v11

_Originally posted by @kazupon in https://github.com/intlify/vue-i18n/issues/2170#issuecomment-2833370761_

PR #2177, opened 2025-04-30
feat: Part options support $d

mauryapari:
Part option support for d.

Would resolve half of https://github.com/intlify/vue-i18n/issues/2171


packages/vue-i18n-core/test/composer.test.ts
```diff
@@ -1058,6 +1058,83 @@ describe('d', () => {
       '12/20/2012, 07:00 AM'
     )
   })
+
+  test.only('parts formatting', () => {
```
kazupon, line 1062:
```suggestion
  test('parts formatting', () => {
```

kazupon (approved):
Thank you! 
LGTM!

PR #2193, opened 2025-05-26
fix(typing): n() & d() output depending "part" option

### LINKED ISSUE #2192: Regression with "Part options support for $n"
### Reporting a bug?

#2170
#2174
The 2 PR **change typing** of `n()` from simple `string` to `string | Intl.NumberFormatPart[]`.
Now it's necessary to **complexify all usages** of `n()`: we have to check type and cast or convert data...

### Expected behavior

/

### Reproduction

```html
<MyComponent :myPropsOfTypeString="n(42)" />
```

### System Info

```shell
/
```

### Screenshot

_No response_

### Additional context

_No response_

### Validations

pinguet62:
fix #2192

Hello 🖖

I don't know how to fix 2 failings tests
* `expectTypeOf( d<Date, string>(..., {..., part: true }) ).toEqualTypeOf<Intl.DateTimeFormatPart[]>()`
* `expectTypeOf( n<string>(..., {..., part: true }) ).toEqualTypeOf<Intl.NumberFormatPart[]>()`

Probably, declare the 2nd generic type as `string` is in conflict with `{ part: true }` *type inference*.
The correct usage should be *type inference*, without generic declaration.
So in my opinion, like these tests was added in PR introducing regression, they are not correct, and can be removed.


kazupon:
Welcome contribution!
I'm happy to your contributrion!
But, This PR has unit test errors. Could you resolve it? 🙏

PR #2204, opened 2025-06-17
fix: error on duplicate useI18n calling on local scope

kazupon:
This PR is bringed up from #2203

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling to prevent duplicate calls to `useI18n` within the same component, displaying a clear error message if this occurs.

- **Tests**
  - Added a test case to ensure an error is thrown when `useI18n` is called multiple times in the same local scope.

### DIFF
--- a/packages/vue-i18n-core/src/errors.ts
+++ b/packages/vue-i18n-core/src/errors.ts
@@ -22,7 +22,9 @@ export const I18nErrorCodes = {
   CANNOT_SETUP_VUE_DEVTOOLS_PLUGIN: 30,
   NOT_INSTALLED_WITH_PROVIDE: 31,
   // unexpected error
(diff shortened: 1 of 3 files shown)


AbdallahAlhaddad:
This PR causes some inconvience as we sometimes want to run the `useI18n` composable more than once. see the following usecase.

```ts
// composableThatUsesI18n.ts
export function composableThatUsesI18n() {
  const { locale } = useI18n()
  // ...
}
```

```vue
<script setup lang="ts">
// component.vue
const { t } = useI18n({ useScope: 'local' })
composableThatUsesI18n() //  👈 Triggers duplicate `useI18n` calling error
</script>
```
## Suggested solution
Could we handle this silently inside the useI18n Composable?

PR #2208, opened 2025-06-22
fix: declaration order in Number formatting with options ResourceKeys

kazupon:
back port from #2205


kazupon:
@coderabbitai review

Issue #2201, opened 2025-06-10
Minor bug in "Getting started" StackBlitz example code


becseya:
### Reporting a bug?

[Getting started docs](https://vue-i18n.intlify.dev/guide/essentials/started.html) links to [stackblitz](https://stackblitz.com/edit/vue-i18n-get-started), where the code contains a typo (_mesasge_ rather _message_) causing a bug described below.

Not familiar with the file format, but suspecting the same typo at line 7 in [spec/syntax.ebnf](https://github.com/intlify/vue-i18n/blob/d4c6e3a3c8b21fc3810c0fafd6ed51135bf0645b/spec/syntax.ebnf#L7)

Can't add PR, as stackblitz code is not tracked.

### Expected behavior

When setting `locale: 'en'` in the code editor at line 6 in __main.js__, I'm expecting the preview to change to _"hello world"_ but it changes to _"message.hello"_

### Reproduction

Reproduction is straight forward based upon text and description.

### System Info

```shell
System not relevant to issue.
```

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/intlify/vue-i18n/blob/master/.github/CONTRIBUTING.md)
- [x] Read the [Documentation](https://vue-i18n.intlify.dev/)
- [x] Check that there isn't [already an issue](https://github.com/intlify/vue-i18n/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussions](https://github.com/intlify/vue-i18n/discussions)

babu-ch:
Probably the same

https://github.com/intlify/vue-i18n/issues/2046

PR #2209, opened 2025-06-22
Fix typos in syntax specification file

babu-ch:
ref 
https://github.com/intlify/vue-i18n/issues/2201

### About the StackBlitz typo fix

I could fork the StackBlitz code and update the link.
That way, it wouldn't require any effort on your side, but I think it would be better if the fix comes from @kazupon .

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected typographical errors in grammar comments and rules for improved clarity.

### DIFF
--- a/spec/syntax.ebnf
+++ b/spec/syntax.ebnf
@@ -1,10 +1,10 @@
 (*
- * Inltify message syntax v0.5.0
+ * Intlify message syntax v0.5.0
  * (vue-i18n compatible)
  *)
 
 (* entrypoint *)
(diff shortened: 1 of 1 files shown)


kazupon (approved):
Good catch!
Thanks!

kazupon:
> I could fork the StackBlitz code and update the link.
> That way, it wouldn't require any effort on your side, but I think it would be better if the fix comes from @kazupon .

Thanks!
I hope that you would fork and update it.

PR #2221, opened 2025-07-02
fix: typo

kazupon:
Thanks to AI, we have just found out the typo.
we have not noticed the typo for 5 years, so far 😂 

I noticed this while working on a side project.
https://github.com/kazupon/gunshi/pull/200/commits/08111f8200cfe8c80d54cdb7ed45b2f30e0d57b8#diff-dc06d3cebbf88f0f86a5b8eefba8c4a2e349130eec7d1af4b69ecfbe3ffb909cR377

### DIFF
--- a/packages/core-base/src/context.ts
+++ b/packages/core-base/src/context.ts
@@ -335,7 +335,7 @@ export interface CoreInternalContext {
  */
 export const VERSION = __VERSION__
 
-export const NOT_REOSLVED = -1
+export const NOT_RESOLVED = -1
 
 export const DEFAULT_LOCALE = 'en-US'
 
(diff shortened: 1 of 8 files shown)


maxdzin:
@kazupon It is nice that it was fixed, but it caused this issue:
https://github.com/intlify/h3/issues/50
which suddenly breaks work of other intlify packages.

hannoeru:
@kazupon This looks like a breaking changes, other package that using core-base will also need to update the imports and limit the version of core-base to ^11.1.8. or we can have a alias for the old consistent to make sure backward compatibility.

I just notice that when I do package update, I got this error:
```
Uncaught SyntaxError: The requested module '/_nuxt/@fs/Users/hanlee/projects/bx-console/node_modules/.cache/vite/client/deps/@intlify_core-base.js?v=8c3f083d' does not provide an export named 'NOT_REOSLVED' (at vue-i18n.mjs:7:282)
```

kazupon:
Sorry, I overlooked that point.
v11 will be reverted.

PR #2400, opened 2026-02-23
feat!: Support automatic pluralization with `Intl.PluralRules`

### LINKED ISSUE #2020: Add Pluralization rules based on Intl.PluralRules
### Clear and concise description of the problem

The default pluralization rules are pretty simple and Intl.PluralRules has [97% browsersupport](https://caniuse.com/?search=Intl.PluralRules) so it would be nice to have an implementation based on it with fallback to the current simple one. Which maybe needs revising since it can currently return a number greater than choicesLength - 1 that then crashes vue

### Suggested solution

Here is an implementation that works with the current api but the implementation could be cleaner without the Proxy

```typescript
const pluralOrder: Intl.LDMLPluralRule[] = [
  'zero',
  'one',
  'two',
  'few',
  'many',
  'other',
]

function pluralSort(a: Intl.LDMLPluralRule, b: Intl.LDMLPluralRule) {
  return pluralOrder.indexOf(a) - pluralOrder.indexOf(b)
}

const pluralRules = new Proxy(
  {} as {
    [lang: string]:
      | ((choice: number, choicesLength: number) => number)
      | undefined
  },
  {
    get: (target, prop: string) => {
      if (prop in target) return target[prop]

      try {
        const rules = new Intl.PluralRules(prop)
        const categories = rules
          .resolvedOptions()
          .pluralCategories.sort(pluralSort)

        return (target[prop] = (choice: number, choicesLength: number) => {
          const rule = rules.select(choice)
          const n = categories.indexOf(rule)

          return Math.min(n, choicesLength - 1)
        })
      } catch (e) {
        target[prop] = undefined
      }
    },
  },
)
```

### Alternative

If shipping it by default is considered too large it could live in an extra import so that it's easy to use
```typescript
import pluralRules from '@intlify/plural'
```

The implementation could also be cleaner if instead of an object pluralRules could be a function that receives the locale maybe it would be a factory so that it only gets called once per locale

### Additional context

_No response_

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/intlify/vue-i18n/blob/master/.github/CONTRIBUTING.md)
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

kazupon:
resolve #2020 

## Summary by CodeRabbit

* **New Features**
  * Implemented automatic pluralization using Intl.PluralRules for multi-language support with CLDR-compliant plural categories.

* **Documentation**
  * Updated guides with Vue 3 syntax improvements and clearer examples.
  * Added comprehensive migration guidance for pluralization behavior in v12.
  * Translated documentation updates across English, Japanese, and Chinese.

* **Tests**
  * Added test coverage for new Intl.PluralRules-based pluralization behavior.

* **Chores**
  * Updated ignore patterns.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,10 @@ docs/jp/api/vue
 docs/jp/api/index.md
 !docs/api/typedoc-sidebar.json
 .notes
+.data
+.playwright-cli
+.report
+.claude
 
 *.log
 *.swp
--- a/docs/guide/advanced/component.md
+++ b/docs/guide/advanced/component.md
@@ -51,8 +51,15 @@ const url = ref('/term')
 <template>
   <div id="app">
     <!-- ... -->
-    <i18n-t keypath="term" tag="label" for="tos">
-      <a :href="url" target="_blank">{{ t('tos') }}</a>
+    <i18n-t
+      keypath="term"
+      tag="label"
+      for="tos"
+    >
+      <a
+        :href="url"
+        target="_blank"
+      >{{ t('tos') }}</a>
     </i18n-t>
     <!-- ... -->
   </div>
@@ -128,11 +135,14 @@ const refundLimit = ref(30)
 <template>
   <div id="app">
     <!-- ... -->
-    <i18n-t keypath="info" tag="p">
-      <template v-slot:limit>
+    <i18n-t
+      keypath="info"
+      tag="p"
+    >
+      <template #limit>
         <span>{{ changeLimit }}</span>
       </template>
-      <template v-slot:action>
+      <template #action>
         <a :href="changeUrl">{{ t('change') }}</a>
       </template>
     </i18n-t>
--- a/docs/guide/essentials/local.md
+++ b/docs/guide/essentials/local.md
@@ -157,7 +157,9 @@ const { t } = useI18n({
       <p>This is good service</p>
     </div>
     <div class="footer">
-      <button type="button">{{ t('buttons.save') }}</button>
+      <button type="button">
+        {{ t('buttons.save') }}
+      </button>
     </div>
   </div>
 </template>
--- a/docs/guide/essentials/pluralization.md
+++ b/docs/guide/essentials/pluralization.md
@@ -114,11 +114,44 @@ As result the below:
 <p>too many bananas</p>
 ```
 
+## Automatic Pluralization with `Intl.PluralRules`
+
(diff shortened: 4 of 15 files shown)


absidue:
Will there be a way to opt-out of this breaking change or is the recommended approach to stay on v11 even after v12 is released? Introducing unintuitive, language specific behaviour that expects translators to look things up in a completely unrelated source, means that using plurals with vue-i18n will become unfeasible for open source projects with volunteer submitted translations.

It is already an issue that at any given moment at least 3 strings fail to compile because of translated or missing placeholders, the new magic behaviour will make things a lot worse. This breaking change is making me reconsider whether it is even worth having plural strings, with the significant maintenance burden that this change will bring. Additionally more magic framework specific logic and behaviour makes adding translations even more complicated which will discourage people from submitting translations to projects.

kazupon:
Thank you for your feedback!

> Will there be a way to opt-out of this breaking change or is the recommended approach to stay on v11 even after v12 is released?

v11 will continue to be maintained for the foreseeable future. I will paln to release v12 for Vue 3.6 release
I will plan to backport from v12 to v11 as needed.

> It is already an issue that at any given moment at least 3 strings fail to compile because of translated or missing placeholders, the new magic behaviour will make things a lot worse. This breaking change is making me reconsider whether it is even worth having plural strings, with the significant maintenance burden that this change will bring. Additionally more magic framework specific logic and behaviour makes adding translations even more complicated which will discourage people from submitting translations to projects.

I think your concern is not an issue with the PR #2400 implementation, for the following reasons:

1. Custom pluralRules take highest priority: If users define their own plural rules, those are always used first. Existing custom rules continue to work exactly as before.
2. Fallback based on message form count: When the number of pipe-separated forms in a message exceeds the locale's CLDR category count, vue-i18n
automatically falls back to the previous English-based default rule (pluralDefault). This means existing translations are never broken by the new behavior.
3. Environments without Intl.PluralRules: The previous default rule (pluralDefault) is used as-is.

For example, an existing Russian message like `car: '{n} машина | {n} машин'` (2 forms) has fewer forms than Russian's CLDR categories (one, few, many, other = 4), so Intl.PluralRules is correctly applied and selects the best match. Conversely, if a message has more forms than the locale's category count, it gracefully falls back to the previous behavior. In either case, existing translations will not break.

absidue:
Thank you for your elaborate answer. it is great to hear that backwards compatibility is built-in and that vue-i18n will handle it transparently. In that case it doesn't sound like there will be too many blockers on us migrating to v12. Thank you for all the work yoi do on vue-i18n!

PR #2441, opened 2026-03-11
fix: mark several packages as having side effects

sumimakito:
This pull request sets `sideEffects` to `true` for the following packages:

- @intlify/core-base
- @intlify/core
- @intlify/vue-i18n-core
- @intlify/vue-i18n

… because they have side effect calls: `registerMessageCompiler`, `registerMessageResolver`, `registerLocaleFallbacker`, and `initFeatureFlags` module scope. With `sideEffects: false`, bundlers may tree-shake these entry points away when only specific exports are imported. This may silently break the library at runtime.

I'm unsure if this was intended. Please feel free to leave comments for discussion.

## Summary by CodeRabbit

* **Chores**
  * Updated build configuration across core packages to optimize module bundling behavior.

### DIFF
--- a/packages/core-base/package.json
+++ b/packages/core-base/package.json
@@ -31,7 +31,7 @@
     "node": ">= 16"
   },
   "type": "module",
-  "sideEffects": false,
+  "sideEffects": true,
   "files": [
     "dist"
   ],
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -31,7 +31,7 @@
     "node": ">= 16"
   },
   "type": "module",
-  "sideEffects": false,
+  "sideEffects": true,
   "files": [
     "dist"
   ],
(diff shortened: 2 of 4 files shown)


kazupon:
Thank you for this PR!

This PR made me realize that the `sideEffects` field in several packages was not correctly configured I hadn't properly accounted for the `register*` calls and other module-scope side effects.

hat said, I'd like to keep `sideEffects: false` for the `vue-i18n` package to minimize bundle size.

Looking into this, I think the right fix is to move the:

- `registerMessageCompiler()`
- `registerMessageResolver()`
- `registerLocaleFallbacker()`

These does call into `createI18n()` instead of running them at module scope. Since users always go through `createI18n()`, this should be safe.

I'll work on that refactoring.

Thanks again for bringing this to my attention! :)

sumimakito:
@kazupon Thank you for the prompt response! Your fix plan sounds more ideal. I'm looking forward to seeing these issues fixed in the new releases, but please take your time. vue-i18n is a useful project.

sumimakito:
I'm closing this PR as there will be a more proper fix for this issue.

PR #2463, opened 2026-04-07
perf: gate performance.mark/measure behind app.config.performance

### LINKED ISSUE #2462: performance.mark/measure called unconditionally in dev mode causes significant overhead with many components
## Description

In development mode, `@intlify/core-base` unconditionally calls `performance.mark()` and `performance.measure()` on every `t()` invocation. Each translation triggers 3 sets of mark/measure:

- `intlify message resolve`
- `intlify message compilation`
- `intlify message evaluation`

This is initialized in `@intlify/shared` with only a `process.env.NODE_ENV !== 'production'` guard — there is no opt-in flag like Vue's `app.config.performance`:

```js
// @intlify/shared/dist/shared.mjs
if ((process.env.NODE_ENV !== 'production')) {
    const perf = inBrowser && window.performance;
    if (perf && perf.mark && perf.measure && perf.clearMarks && perf.clearMeasures) {
        mark = (tag) => { perf.mark(tag) }
        measure = (name, startTag, endTag) => {
            perf.measure(name, startTag, endTag)
            perf.clearMarks(startTag)
            perf.clearMarks(endTag)
        }
    }
}
```

Image

## Impact

In our application (a node-based workflow editor), switching between workflows renders ~160 Vue components simultaneously, each with multiple `t()` calls. Chrome DevTools profiling shows **~240ms** spent on `mark` (98ms) + `measure` (141.7ms) during a single workflow switch — roughly 20% of the total frame time.

This overhead only exists in dev mode and does not affect production builds, but it distorts performance profiling results and slows down development iteration.

## Suggestion

Consider gating `performance.mark/measure` behind an opt-in flag (similar to Vue's `app.config.performance`), or making them conditional on Vue DevTools actually being connected. The `__v_emitter` events already have their own guard (`if (emitter && start)`), so the mark/measure calls could follow the same pattern.

## Environment

- vue-i18n: 9.14.5
- @intlify/core-base: 9.14.5
- Browser: Chrome 136
- Bundler: Vite 6 (dev server)

mvanhorn:
## Summary

Dev mode calls `performance.mark()` and `performance.measure()` on every `t()` invocation unconditionally. This adds ~240ms overhead in apps with many components. This PR gates those calls behind `app.config.performance`, matching Vue core's opt-in behavior.

## Why this matters

Profiling a workflow editor with ~160 components showed 98ms in `mark` + 142ms in `measure` per workflow switch. That's 20% of the total frame time, all from instrumentation nobody asked for. Wrapping the initialization in `initPerformanceMeasurement()` and calling it only when `app.config.performance` is true eliminates this cost for the common case.

## Changes

- `packages/shared/src/utils.ts`: Moved `mark`/`measure` initialization into a new `initPerformanceMeasurement()` function. The functions stay `undefined` until explicitly enabled, so existing `if (mark && measure)` guards in `translate.ts` skip them by default.
- `packages/vue-i18n-core/src/i18n.ts`: Import and call `initPerformanceMeasurement()` in the plugin `install()` hook when `__DEV__ && app.config.performance` is true.

## Testing

- Without `app.config.performance = true`: `mark` and `measure` remain undefined, zero overhead from Performance API calls in dev mode.
- With `app.config.performance = true`: behavior is identical to the current dev mode (all three measurement phases fire).

Fixes #2462

This contribution was developed with AI assistance (Claude Code).

## Refactor
* Performance measurement initialization now occurs on-demand during plugin setup instead of at module load time, improving initialization efficiency.

### DIFF
--- a/packages/shared/src/utils.ts
+++ b/packages/shared/src/utils.ts
@@ -8,24 +8,26 @@ export const inBrowser: boolean = typeof window !== 'undefined'
 export let mark: (tag: string) => void | undefined
 export let measure: (name: string, startTag: string, endTag: string) => void | undefined
 
-if (__DEV__) {
-  const perf = inBrowser && window.performance
+export function initPerformanceMeasurement(): void {
+  if (__DEV__) {
+    const perf = inBrowser && window.performance
 
-  if (
-    perf &&
-    perf.mark &&
-    perf.measure &&
-    perf.clearMarks &&
-    // @ts-ignore browser compat
-    perf.clearMeasures
-  ) {
-    mark = (tag: string): void => {
-      perf.mark(tag)
-    }
-    measure = (name: string, startTag: string, endTag: string): void => {
-      perf.measure(name, startTag, endTag)
-      perf.clearMarks(startTag)
-      perf.clearMarks(endTag)
+    if (
+      perf &&
+      perf.mark &&
+      perf.measure &&
+      perf.clearMarks &&
+      // @ts-ignore browser compat
+      perf.clearMeasures
+    ) {
+      mark = (tag: string): void => {
+        perf.mark(tag)
+      }
+      measure = (name: string, startTag: string, endTag: string): void => {
+        perf.measure(name, startTag, endTag)
+        perf.clearMarks(startTag)
+        perf.clearMarks(endTag)
+      }
     }
   }
 }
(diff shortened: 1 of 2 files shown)


kazupon:
Thank you for taking the time to work on this and send the fix!

However, after looking into it, I found that this change does not address the underlying issue. I have handled the root cause on my side instead at https://github.com/intlify/vue-i18n/pull/2465

Thanks!

PR #2515, opened 2026-05-27
fix(shared): clear performance measures to prevent User Timing API me…

valeriipcorefy:
## Bug Description

The `measure` utility in `@intlify/shared` calls `perf.clearMarks()` for
start/end tags but never calls `perf.clearMeasures()`. This causes
`PerformanceMeasure` entries to accumulate indefinitely in the browser's
performance timeline buffer during development.

In long-running sessions with frequent translations, the buffer grows to
hundreds of thousands of entries, degrading `clearMarks()` and GC
operations to O(N) and consuming significant CPU time (~28% in our case).

## Fix

Add `perf.clearMeasures(name)` immediately after recording each measure
in `packages/shared/src/utils.ts`. This clears the entry from the live
buffer while preserving data already captured by an active DevTools
Performance recording.

## Tests

Added `packages/shared/test/performance.test.ts` — runs in JSDOM
environment, verifies that `mark` and `measure` correctly delegate to
the Performance API and that all marks and measures are cleaned up.

All existing tests pass (`pnpm test:unit` — 811 tests, 0 failures).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved performance measurement cleanup to ensure measurements are properly cleared after being recorded, preventing potential memory issues and ensuring accurate subsequent measurements.

* **Tests**
  * Added comprehensive test suite for performance timing utilities to verify correct operation and ensure proper cleanup of all performance tracking elements in supported environments.

[Review Change Stack](https://app.coderabbit.ai/change-stack/intlify/vue-i18n/pull/2515?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/shared/src/utils.ts
+++ b/packages/shared/src/utils.ts
@@ -28,6 +28,7 @@ if (__DEV__) {
       perf.measure(name, startTag, endTag)
       perf.clearMarks(startTag)
       perf.clearMarks(endTag)
+      perf.clearMeasures(name)
     }
   }
 }
--- a/packages/shared/test/performance.test.ts
+++ b/packages/shared/test/performance.test.ts
@@ -1,0 +1,57 @@
+// @vitest-environment jsdom
+import { afterEach, describe, expect, it, vi } from 'vitest'
+import { mark, measure } from '../src/utils'
+
+describe('performance timing utils', () => {
+  afterEach(() => {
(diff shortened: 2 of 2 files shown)


kazupon (approved):
Thanks!
LGTM!

Issue #324, opened 2021-01-26
Incorrect generated types & fix for "eslint@typescript-eslint/unbound-method" error


IlCallo:
### vue & vue-i18n version
ex:
- vue: 3.0.5
- vue-i18n: 9.0.0-rc.4

### What is actually happening?
Hi there!
I'm testing the new package version and I got stuck for a while on a couple of problems related to the generated types:
- `declare` keyword used on `type` and `interface` exports, which has no effect (AFAIK)
- declaration of `useI18n` functions using method style (`t(): string`) instead of property style (`t: () => string`), which causes a "eslint@typescript-eslint/unbound-method" linting error when destructured

The first problem shouldn't have any side effect, it's just pointless: `declare` keyword has no effect when used in that way. You should probably remove those anyway, just to be sure.

The second one would force anyone using TypeScript linter to disable an error every time the composable is used.
I guess you defined those using the method notation to be able to add overloads, but TS assume that functions on object defined with method notation MIGHT be using the local context, that's why the linter then complains about this.
You can accomplish the same thing by declaring those functions and overloads at root level, then adding a property with the function name and with type as the `typeof` of the declared function.

Example: 
```ts
declare function t(key: Path | number): string;
declare  function t(key: Path | number, plural: number, options?: TranslateOptions): string;
declare function t(key: Path | number, defaultMsg: string, options?: TranslateOptions): string;
// Other overloads

export interface Composer<Messages = {}, DateTimeFormats = {}, NumberFormats = {}, Message = VueMessageType> {
    // other properties
    t: typeof t;
    // other properties
}
```

I tried this out on-the-fly and it seems to work correctly.
Other Vue3-ready packages I used doesn't seem to share this problem

kazupon:
Hmm, 🤔 
The `*.d.ts` of  vue-i18n-next is generated with api-extractor.
https://github.com/microsoft/rushstack/tree/master/apps/api-extractor

I think that api-extractor have problems.

IlCallo:
I'm not familiar with API Extractor (yet).
Dunno how you can solve the first problem, but I the second one is probably due to how you define the composer interface in https://github.com/intlify/vue-i18n-next/blob/master/packages/vue-i18n/src/composer.ts

If you extract `t` (and other functions) as root level definitions, in a similar fashion to what I suggested into the first post, it's probable that API Extractor will be able to generate typings correctly and fix the second error (which right now is the one actually causing real problems)

kazupon:
[...]

PR #2531, opened 2026-07-14
Mark Composer te as context-independent

NVolcz:
`Composer.te` is implemented as a context-independent closure, but its type declaration does not express that. As a result, safely destructuring it triggers `@typescript-eslint/unbound-method`:

```ts
const { te } = useI18n()
te('some.key')
```

This adds this: void to the te declaration so its type matches the runtime implementation. It also adds a type test for the context-independent contract. There are no runtime changes.

Related:
- https://github.com/intlify/vue-i18n/issues/324
- https://typescript-eslint.io/rules/unbound-method/

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TypeScript type checking for the translation-existence method when called or passed as a callback.

* **Tests**
  * Added type coverage to verify the method can be used without an implicit `this` context.


kazupon (approved):
Thanks!
LGTM!

PR #2559, opened 2026-07-28
fix(core-base): use requested locale for Intl when formats fall back

karladler:
## Summary

- When named datetime/number formats are resolved through the fallback chain (e.g. locale `en-GB` with formats under `en`), `resolveFormatLocale` correctly returns the **resource key** for looking up format options, but that same key was also passed to `Intl.DateTimeFormat` / `Intl.NumberFormat`.
- As a result, `d(date, 'short')` with `locale: 'en-GB'` and `datetimeFormats.en` formatted as US order (`01/31/2026`) instead of GB order (`31/01/2026`).
- This change keeps format options from the fallback resource locale, but always passes the **requested** locale (with `!` stripped) to Intl. Formatter cache keys now include both locales so caches stay correct when clearing formats.

## Before / after

```ts
createI18n({
  legacy: false,
  locale: 'en-GB',
  datetimeFormats: {
    en: { short: { year: 'numeric', month: '2-digit', day: '2-digit', timeZone: 'UTC' } },
  },
})

d(new Date('2026-01-31T00:00:00Z'), 'short')
// before: 01/31/2026
// after:  31/01/2026
```

## Test plan

- [x] Added core-base datetime/number tests for region locale + language-code format resources (`en-GB`/`en`, `de-CH`/`de`)
- [x] Updated existing cross-locale fallback expectations (requested locale is used for Intl; borrowed options only)
- [x] Updated formatter cache key assertions / clear behavior
- [x] `pnpm test:unit` passes locally
- [x] Verified with local repro (unit + Playwright) against this branch

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved date and number formatting when a regional locale falls back to its language code.
  * Ensured the requested regional locale is still used for final `Intl` formatting.
  * Improved formatter cache separation and reuse across locales and formatting options.

* **Tests**
  * Added coverage for regional locale fallback scenarios.
  * Updated date formatting expectations to reflect corrected localized output.

### DIFF
--- a/packages/core-base/src/datetime.ts
+++ b/packages/core-base/src/datetime.ts
@@ -189,7 +189,7 @@ export function datetime<Context extends CoreContext<Message, {}, {}, {}>, Messa
     return !part ? formatter.format(value) : formatter.formatToParts(value)
   }
 
-  const targetLocale = resolveFormatLocale(
+  const formatLocale = resolveFormatLocale(
     context,
     key,
     locale,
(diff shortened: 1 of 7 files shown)


kazupon:
Thank you for working on this!

I understand the intention: when `en-GB` falls back to a format definition stored under `en`, `en-GB` should still be passed to `Intl.DateTimeFormat`.

However, the current implementation applies not only to implicit fallback within the same language, but also to explicit fallback to a different language via `fallbackLocale`. In fact, the existing `en-US -> ja-JP` test changes from Japanese-style output to US-style output. The same change in behavior also applies to `Intl.NumberFormat`.

Although this PR does not directly change message fallback, messages continue to use the fallback `targetLocale` for compilation and pluralization. As a result, the meaning of fallback would differ between messages and Intl formats:

- Messages are evaluated using the fallback locale.
- Datetime and number formats borrow only the options from the fallback locale and are evaluated using the requested locale.

In addition, Vue I18n's `Locale` type permits arbitrary strings. For example, a custom locale such as `en_US`, which is invalid for Intl, may previously have worked by falling back to a valid locale such as `en-US`. Passing the requested locale directly to Intl could now cause a `RangeError` in such cases.

Could you clarify which of the following behaviors is intended?

1. Preserve the requested locale only for implicit fallback within the same language, such as `en-GB -> en` or `de-CH -> de`.
2. Borrow only the format options for every fallback, including explicit fallback to a different language.

I think limiting this change to implicit fallback within the same language would be safer and would minimize its impact on existing behavior. If the intention is to change every fallback case, I believe this should be treated as a compatibility-affecting behavioral change and accompanied by documentation and additional tests.

PR #2564, opened 2026-08-14
fix(core-base): invalidate locale chain cache when fallback changes

### LINKED ISSUE #2562: fallbackWithLocaleChain caches the chain by start locale only, so a changed fallbackLocale is ignored
I'm building a Python port of this library [here](https://github.com/velis74/vue-i18n-python) — a faithful 
one, which means reading `core-base` closely and recording what it actually does rather than what I
assume it does — and I have found this.

`fallbackWithLocaleChain` memoises the computed chain in `context.__localeChainCache`, keyed by the
**start locale alone** (`packages/core-base/src/fallbacker.ts`, lines 135 and 160). The
`fallbackLocale` argument is not part of the key, so a second call with the same start locale and a
*different* `fallbackLocale` receives the first call's answer.

Reproduction, on `@intlify/core-base@11.4.8`, Node 26:

```js
import { fallbackWithLocaleChain } from '@intlify/core-base'

const ctx = {}
fallbackWithLocaleChain(ctx, { sl: ['en'], default: ['en'] }, 'sl')        // ["sl","en"]  ✅
fallbackWithLocaleChain(ctx, { sl: ['de', 'hr'], default: ['de'] }, 'sl')  // ["sl","en"]  ❌
fallbackWithLocaleChain({},  { sl: ['de', 'hr'], default: ['de'] }, 'sl')  // ["sl","de","hr"]
```

The third line shows the correct answer; the second is the stale one.

**Expected:** the chain reflects the `fallbackLocale` that was passed.
**Actual:** it reflects whichever `fallbackLocale` happened to be passed first for that start locale,
for the lifetime of the context.

This is reachable from application code whenever `fallbackLocale` changes at runtime — a language
picker that also changes the fallback order, or a lazily loaded locale set. `updateFallbackLocale`
clears the cache and is presumably why this has not bitten more often, but that puts the correctness
of the cache on the caller remembering to route through a specific function, and
`i18n.global.fallbackLocale = ...` does not.

Two possible fixes, depending on what you consider the contract: include the `fallbackLocale` in the
cache key (a stable stringification, as `generateFormatCacheKey` already does for message formats),
or clear the cache from the `fallbackLocale` setter so the existing key stays valid.

Happy to open a PR for either if you have a preference.

haoku123:
## Description

Fixes #2562: `fallbackWithLocaleChain` cached the resolved chain keyed only by the start locale. A subsequent call with the **same start locale but a different fallback locale** returned the stale chain computed earlier.

## Root Cause

`__localeChainCache` was a `Map<Locale, Locale[]>`. Once `('ja', 'en')` populated the `ja` entry, calling `('ja', 'fr')` would still hit the cached `['ja', 'en']` chain even though the fallback had changed to `fr`.

## Changes

- `packages/core-base/src/context.ts`: cache value type widened to `{ fallback: FallbackLocale; chain: Locale[] }`
- `packages/core-base/src/fallbacker.ts`: only reuse the cached entry when the fallback **reference** matches; otherwise recompute and refresh the entry

## Tests

Added a `cache invalidation` describe block in `fallbacker.test.ts`:
- same start locale, different fallback → chain recomputed
- same start locale, same fallback reference → cached chain reused (referential identity preserved)

`vitest run test/fallbacker.test.ts`: 42/42 pass.
Control experiment: with the unconditional cache hit restored, the new tests fail with a stale chain, confirming sensitivity.

Note: 3 pre-existing `datetime.test.ts` failures in this repo are timezone-related (local TZ is UTC+8) and unrelated to this change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved locale fallback handling so updated fallback settings are correctly reflected instead of using stale cached results.
  * Preserved efficient reuse of locale chains when fallback settings remain unchanged.
  * Ensured distinct fallback configurations maintain separate cached locale chains.

* **Tests**
  * Added coverage for fallback cache invalidation, in-place fallback updates, distinct configurations, and cache reuse behavior.

### DIFF
--- a/packages/core-base/src/fallbacker.ts
+++ b/packages/core-base/src/fallbacker.ts
@@ -1,4 +1,5 @@
 import {
+  friendlyJSONstringify,
   isArray,
   isBoolean,
   isFunction,
@@ -130,9 +131,16 @@ export function fallbackWithLocaleChain<Message = string>(
     context.__localeChainCache = new Map()
   }
 
-  const cached = context.__localeChainCache.get(startLocale)
-  if (cached && cached.fallback === fallback) {
-    return cached.chain
+  const fallbackKey = friendlyJSONstringify(fallback)
+  let chains = context.__localeChainCache.get(startLocale)
+  if (!chains) {
+    chains = new Map()
+    context.__localeChainCache.set(startLocale, chains)
+  }
+
+  const cached = chains.get(fallbackKey)
+  if (cached) {
+    return cached
   }
 
   const chain: Locale[] = []
@@ -158,7 +166,7 @@ export function fallbackWithLocaleChain<Message = string>(
   if (isArray(block)) {
     appendBlockToChain(chain, block, false)
   }
-  context.__localeChainCache.set(startLocale, { fallback, chain })
+  chains.set(fallbackKey, chain)
 
   return chain
 }
--- a/packages/core-base/src/fallbacker.ts
+++ b/packages/core-base/src/fallbacker.ts
@@ -130,33 +130,35 @@ export function fallbackWithLocaleChain<Message = string>(
     context.__localeChainCache = new Map()
   }
 
-  let chain = context.__localeChainCache.get(startLocale)
-  if (!chain) {
-    chain = []
-
-    // first block defined by start
-    let block: unknown = [start]
-
-    // while any intervening block found
-    while (isArray(block)) {
-      block = appendBlockToChain(chain, block, fallback)
-    }
-
-    // prettier-ignore
-    // last block defined by default
-    const defaults = isArray(fallback) || !isPlainObject(fallback)
-      ? fallback
-      : fallback['default']
-        ? fallback['default']
-        : null
-
-    // convert defaults to array
-    block = isString(defaults) ? [defaults] : defaults
-    if (isArray(block)) {
-      appendBlockToChain(chain, block, false)
-    }
-    context.__localeChainCache.set(startLocale, chain)
+  const cached = context.__localeChainCache.get(startLocale)
+  if (cached && cached.fallback === fallback) {
+    return cached.chain
   }
+
+  const chain: Locale[] = []
+
+  // first block defined by start
+  let block: unknown = [start]
+
+  // while any intervening block found
+  while (isArray(block)) {
+    block = appendBlockToChain(chain, block, fallback)
+  }
+
+  // prettier-ignore
+  // last block defined by default
+  const defaults = isArray(fallback) || !isPlainObject(fallback)
+    ? fallback
+    : fallback['default']
+      ? fallback['default']
+      : null
+
+  // convert defaults to array
+  block = isString(defaults) ? [defaults] : defaults
+  if (isArray(block)) {
+    appendBlockToChain(chain, block, false)
+  }
+  context.__localeChainCache.set(startLocale, { fallback, chain })
 
   return chain
 }
--- a/packages/core-base/test/fallbacker.test.ts
+++ b/packages/core-base/test/fallbacker.test.ts
@@ -250,23 +250,35 @@ describe('fallbackWithLocaleChain', () => {
 
   describe('cache invalidation', () => {
     test('recomputes the chain when the fallback locale changes (#2562)', () => {
-      // First call caches the chain computed with fallback 'en'.
       expect(fallbackWithLocaleChain(ctx, 'en', 'ja')).toEqual(['ja', 'en'])
-
-      // Same start locale, different fallback: the cached entry for 'ja'
-      // must not be reused because it was computed with fallback 'en'.
       expect(fallbackWithLocaleChain(ctx, 'ja', 'ja')).toEqual(['ja'])
+    })
 
-      // Same fallback, different start locale still hits the fresh entry.
-      expect(fallbackWithLocaleChain(ctx, 'ja', 'en')).toEqual(['en', 'ja'])
+    test('recomputes the chain for different fallback objects with the same start locale (#2562)', () => {
+      expect(fallbackWithLocaleChain(ctx, { sl: ['en'], default: ['en'] }, 'sl')).toEqual([
+        'sl',
+        'en'
+      ])
+
+      expect(fallbackWithLocaleChain(ctx, { sl: ['de', 'hr'], default: ['de'] }, 'sl')).toEqual([
+        'sl',
+        'de',
+        'hr'
+      ])
+    })
+
+    test('recomputes the chain when the fallback is mutated in place', () => {
+      const fallback: Record<string, string[]> = { sl: ['en'], default: ['en'] }
+      expect(fallbackWithLocaleChain(ctx, fallback, 'sl')).toEqual(['sl', 'en'])
+
+      fallback.sl = ['de']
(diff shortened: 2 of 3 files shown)


packages/core-base/src/fallbacker.ts
```diff
@@ -130,33 +130,35 @@ export function fallbackWithLocaleChain<Message = string>(
     context.__localeChainCache = new Map()
   }
 
-  let chain = context.__localeChainCache.get(startLocale)
-  if (!chain) {
-    chain = []
+  const cached = context.__localeChainCache.get(startLocale)
+  if (cached && cached.fallback === fallback) {
```
kazupon: **[suggestion]** Cache reuse is `cached.fallback === fallback`. That fixes #2562 when the caller passes a different fallback reference (string `'en'` vs `'ja'`, or two object literals). It does not fix the same start locale when the fallback *contents* change but the reference does not (`fallback.sl = ['de']`, `fallback.push('ja')`, or mutating `default`). That path still returns the stale chain. It was already broken before this PR, and `updateFallbackLocale` remains the only way to invalidate it.

This matters for the actual cache contract:
- Composer setter (`composer.ts` ~2114–2117) already assigns a new value and calls `updateFallbackLocale`, so the common Vue I18n assignment path does not need this check.
- `updateFallbackLocale` is `@internal` and unused by `createCoreContext` consumers that write `ctx.fallbackLocale = next` or call `fallbackWithLocaleChain` directly. Those callers are exactly #2562. Replacement is now safe; in-place mutation is not.
- After the Composer setter, `_context.fallbackLocale` is the raw assigned object while `_fallbackLocale.value` is Vue's reactive proxy of it. `t()` / `translate` pass the raw value; `te()` / `tm()` pass the proxy. `===` is false, so those two call sites will keep missing and overwriting each other's cache entry. Results stay correct (same contents); it is wasted recomputation and shows how brittle reference identity is next to Vue wrapping.

#2562's first proposal (stable stringification of `fallbackLocale`, same idea as `generateFormatCacheKey`) would treat value identity, cover in-place mutation, and avoid the proxy/raw split. The third approach here is cheaper and enough for the issue repro, but it is not a complete self-contained cache key.

**Suggestion:** Decide the contract explicitly. If the cache should be correct without `updateFallbackLocale` whenever fallback *contents* change, key by a stable serialization of `fallback` (and keep one entry per start locale, or include the serialized fallback in the Map key). If reference equality is intentional, document that mutating an array/object fallback in place still requires `updateFallbackLocale` (or a new assignment), and consider comparing with a small structural equality helper so Composer `t` / `te` do not thrash the cache after a setter assignment.

packages/core-base/test/fallbacker.test.ts
```diff
@@ -247,6 +247,29 @@ describe('fallbackWithLocaleChain', () => {
...
+  describe('cache invalidation', () => {
+    test('recomputes the chain when the fallback locale changes (#2562)', () => {
+      // First call caches the chain computed with fallback 'en'.
+      expect(fallbackWithLocaleChain(ctx, 'en', 'ja')).toEqual(['ja', 'en'])
+
+      // Same start locale, different fallback: the cached entry for 'ja'
+      // must not be reused because it was computed with fallback 'en'.
+      expect(fallbackWithLocaleChain(ctx, 'ja', 'ja')).toEqual(['ja'])
+
+      // Same fallback, different start locale still hits the fresh entry.
```
kazupon: **[suggestion]** The new tests prove the string-replacement case and same-reference reuse, but they do not lock the issue's actual repro (two different fallback *objects* for the same start locale). The third assertion is also mis-described: after `fallbackWithLocaleChain(ctx, 'ja', 'ja')`, the cache key for start `'en'` is empty, so `fallbackWithLocaleChain(ctx, 'ja', 'en')` is a miss and a new compute, not a hit on "the fresh entry". The surrounding comments narrate each expect (what the call caches / must not reuse / must reuse) rather than stating a non-obvious constraint.

**Suggestion:** Drop the narrative comments. Replace the third expect with something that actually exercises the cache (e.g. a second call with the same start+fallback is `toBe` the previous array), and add the #2562 object-literal pair (`{ sl: ['en'], default: ['en'] }` then `{ sl: ['de', 'hr'], default: ['de'] }` for start `'sl'`). Optionally add a note-only case that same-reference mutation is still stale if you keep `===`.

kazupon (commented):
Thanks for the fix — this covers the #2562 repro when a different fallback *reference* is passed.

Please also address the two inline comments before merge:

1. `cached.fallback === fallback` only invalidates on reference change. In-place mutation of an array/object fallback still returns a stale chain, and after a Composer setter assign `t` / `te` can miss each other's cache entry (raw value vs Vue proxy). Either key by a stable serialization of `fallback` (as suggested in #2562), or document that in-place mutation still requires `updateFallbackLocale` / a new assignment and avoid the raw-vs-proxy cache thrash.
2. The new tests do not lock the actual #2562 object-literal repro, and the third assertion in the first test is a cache miss (different start locale), not a hit. Please add the object-literal case and drop the narrative comments.

kazupon (changes requested):
I've just reviewed, so please check it out.

packages/core-base/src/fallbacker.ts
haoku123: Addressed in a113cf1: the cache is now keyed per start locale by `friendlyJSONstringify(fallback)` (same approach as `generateFormatCacheKey`), so equal contents hit regardless of reference identity or Vue proxying, and in-place mutation or assignment of different contents recomputes the chain. The raw-vs-proxy thrash between `t()` and `te()` is gone too, since serialization is identity-agnostic.

packages/core-base/test/fallbacker.test.ts
haoku123: Addressed in a113cf1: dropped the narrative comments, added the #2562 object-literal pair (`{ sl: ['en'], default: ['en'] }` then `{ sl: ['de', 'hr'], default: ['de'] }` for start `'sl'`), and added an in-place mutation case (`fallback.sl = ['de']` recomputes). The referential-identity hit test stays (`toBe`), and the mis-described third assertion was removed.

haoku123:
@kazupon thanks for the thorough review! Both points addressed in `a113cf1`:

**Cache key** — switched from `cached.fallback === fallback` to a per-start-locale cache keyed by `friendlyJSONstringify(fallback)` (the same serialization approach as `generateFormatCacheKey`). This means:

- equal contents hit regardless of reference identity, so the raw-vs-Vue-proxy split after a Composer setter no longer thrashes the cache between `t()` and `te()`;
- in-place mutation (`fallback.sl = ['de']`, `fallback.push(...)`) changes the serialized key and recomputes the chain;
- `updateFallbackLocale`'s full reset stays unchanged.

**Tests** — dropped the narrative comments, added the #2562 object-literal repro (`{ sl: ['en'], default: ['en'] }` → `{ sl: ['de', 'hr'], default: ['de'] }` for start `'sl'`), added an in-place mutation case, and kept the referential-identity hit test (`toBe`). The mis-described third assertion was removed.

`packages/core-base/test/fallbacker.test.ts` passes 44/44; the only failing tests in the full core-base run are 3 pre-existing timezone-dependent datetime assertions (also failing on clean `main` in this UTC container).

kazupon (approved):
Thanks!
LGTM!

kazupon:
If you would like to back-port for v11, you can try to contribute!
Thanks!

PR #2565, opened 2026-08-14
fix(core-base): fallbackWithSimple uses map keys and skips default

### LINKED ISSUE #2563: fallbackWithSimple reads a map fallbackLocale's keys instead of its values
I'm building a Python port of this library, which meant porting both fallbackers and testing them
against the real ones — and I have found this. It is low severity and it may well be intentional, so
the report is really a question: is the behaviour wrong, or is the type?

`fallbackWithSimple` (`packages/core-base/src/fallbacker.ts`, lines 87–105) accepts a
`FallbackLocale`, and that type includes the map form:

```ts
export type FallbackLocale = Locale | Locale[] | { [locale in string]: Locale[] } | false
```

For the map case it does this (line 99):

```ts
...(isArray(fallback)
  ? fallback
  : isObject(fallback)
    ? Object.keys(fallback)      // <-- keys, not values
    : ...)
```

So the map's **keys** become the fallback chain. The locales the configuration actually names are
never consulted, and the literal string `"default"` is treated as a locale.

Reproduction, `@intlify/core-base@11.4.8`, Node 26, with
`fallbackLocale = { 'sl-SI': ['sl','en'], sl: ['en'], default: ['en'] }`:

| start | `fallbackWithSimple` | `fallbackWithLocaleChain` |
|---|---|---|
| `sl-SI` | `["sl-SI","sl","default"]` | `["sl-SI","sl","en"]` |
| `sl` | `["sl","sl-SI","default"]` | `["sl","en"]` |
| `de` | `["de","sl-SI","sl","default"]` | `["de","en"]` |

Three things go wrong at once: `"default"` is emitted as a locale name; starting from `de` pulls in
`sl-SI` and `sl`, which the configuration never associated with `de`; and `en` — the only locale the
map actually points at — never appears.

**Why it is mostly invisible, and where it is not.** Every shipped entry point registers
`fallbackWithLocaleChain` (`packages/core/src/index.ts`, `packages/vue-i18n/src/index.ts`,
`packages/petite-vue-i18n/src/index.ts`), so a `vue-i18n` application never reaches this. It is
reached by anyone building directly on `@intlify/core-base` and relying on the default in
`createCoreContext` (line 563, `_fallbacker || fallbackWithSimple`). That is not hypothetical: it is
how my port behaved until I changed its default, and it is why I noticed.

**The question.** If "simple" is meant to be deliberately naive, then the behaviour is fine and the
*signature* is wrong — it should not accept a shape it cannot honour, or the doc comment should say
that a map contributes its keys. If the signature is right, then the implementation should follow
the values, and probably resolve `default` the way `fallbackWithLocaleChain` does.

Either is a small change; I did not want to send a PR before knowing which one you consider correct.

haoku123:
Fixes #2563.

## Problem

`fallbackWithSimple` spreads a map-form `FallbackLocale`'s **keys** into the fallback chain. With `fallbackLocale = { 'sl-SI': ['sl','en'], sl: ['en'], default: ['en'] }`, starting from `sl-SI` produced `['sl-SI','sl','default']` — emitting the literal string `'default'` as a locale name.

## Solution

`fallbackWithSimple` stays naive, as designed: a map contributes its **keys** to the chain (not the values it points at — that is `fallbackWithLocaleChain`'s job). The fix is to omit the `default` key so it is never emitted as a locale name:

```ts
...(isObject(fallback)
  ? Object.keys(fallback).filter(locale => locale !== 'default')
  : ...)
```

Result:

| start | before | after |
| --- | --- | --- |
| `sl-SI` | `['sl-SI','sl','default']` | `['sl-SI','sl']` |
| `sl` | `['sl','sl-SI','default']` | `['sl','sl-SI']` |
| `de` | `['de','sl-SI','sl','default']` | `['de','sl-SI','sl']` |

## Changes

- Filter `default` out of the map-keys chain in `fallbackWithSimple`.
- Restore the existing `{ en: [], ja: [] }` tests as the spec for map handling, plus a case asserting `default` is skipped.
- Update the JSDoc and the hand-maintained API docs (en/jp/zh) to say that a map contributes its keys and `default` is skipped.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fallback locale maps now exclude the `default` key when building fallback chains.
  - Configured locale keys continue to be included as expected.

- **Documentation**
  - Clarified fallback behavior for string, array, and map configurations across English, Japanese, and Chinese documentation.

### DIFF
--- a/docs/api/v11/general.md
+++ b/docs/api/v11/general.md
@@ -521,7 +521,7 @@ export declare function fallbackWithSimple<Message = string>(_ctx: CoreContext<M
 
 A fallback locale function implemented with a simple fallback algorithm.
 
-Basically, it returns the value as specified in the `fallbackLocale` props, and is processed with the fallback inside intlify.
+Basically, the chain consists of `start` plus the specified fallback: for a string or array `fallbackLocale`, the value as specified in the props is used; for a map `fallbackLocale`, the map's **keys** are used and the `default` key is skipped since it is not a locale name.
 
 ### Parameters
 
--- a/docs/jp/api/v11/general.md
+++ b/docs/jp/api/v11/general.md
@@ -521,7 +521,7 @@ export declare function fallbackWithSimple<Message = string>(_ctx: CoreContext<M
 
 単純なフォールバックアルゴリズムで実装されたフォールバックロケール関数。
 
-基本的には、`fallbackLocale` props で指定された値を返し、intlify 内部でフォールバック処理されます。
+基本的には、チェーンは `start` と指定されたフォールバックで構成されます。文字列または配列の `fallbackLocale` の場合、props で指定された値が使用され、マップの `fallbackLocale` の場合、マップの**キー**が使用され、`default` キーはロケール名ではないためスキップされます。
 
 ### パラメータ
 
--- a/docs/zh/api/v11/general.md
+++ b/docs/zh/api/v11/general.md
@@ -521,7 +521,7 @@ export declare function fallbackWithSimple<Message = string>(_ctx: CoreContext<M
 
 使用简单回退算法实现的回退语言环境函数。
 
-基本上，它返回 `fallbackLocale` props 中指定的值，并在 intlify 内部进行回退处理。
+基本上，回退链由 `start` 加上指定的回退组成：对于字符串或数组形式的 `fallbackLocale`，使用 props 中指定的值；对于映射形式的 `fallbackLocale`，使用映射的**键**，并跳过 `default` 键（因为它不是语言环境名称）。
 
 ### 参数
 
--- a/packages/core-base/src/fallbacker.ts
+++ b/packages/core-base/src/fallbacker.ts
@@ -72,9 +72,7 @@ export type LocaleFallbacker = <Message = string>(
  * @remarks
  * A fallback locale function implemented with a simple fallback algorithm.
  *
- * Basically, it returns the value as specified in the `fallbackLocale` props, and is processed with the fallback inside intlify.
- *
- * For a map `fallbackLocale`, the chain is `start` plus the map's **keys**, and the `default` key is skipped since it is not a locale name.
+ * Basically, the chain consists of `start` plus the specified fallback: for a string or array `fallbackLocale`, the value as specified in the props is used; for a map `fallbackLocale`, the map's **keys** are used and the `default` key is skipped since it is not a locale name.
  *
  * @param ctx - A {@link CoreContext | context}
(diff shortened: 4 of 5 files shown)


kazupon:
Thanks for picking this up.

The intended direction is the one posted on #2563: `fallbackWithSimple` stays naive. For a map `fallbackLocale`, the chain is `start` plus the object’s **keys**, not the values. Please do not switch this function to `fallback[start] ?? fallback['default'] ?? []` — that is `fallbackWithLocaleChain`’s job.

What we want instead:

1. Keep using the map keys (the existing `{ en: [], ja: [] }` tests are the spec; please restore them).
2. Omit the `default` key from that list, so it is never emitted as a locale name.
3. Update the JSDoc / API docs to say that a map contributes its keys, and that `default` is skipped.

We are not changing this function to follow map values.

haoku123:
@kazupon understood — reverted the values approach. `fallbackWithSimple` now stays naive:

1. Map keys are used as before (the existing `{ en: [], ja: [] }` tests are the spec, restored unchanged).
2. The `default` key is filtered out of the key list, so it is never emitted as a locale name.
3. JSDoc now states the map contract: the chain is `start` plus the map's keys, with `default` skipped.

Added a regression test (`{ en: [], ja: [], default: ['en'] }` → `['ca', 'en', 'ja']`), and `fallbacker.test.ts` passes 41/41.

kazupon:
Thanks — the implementation and tests now match the intended spec.

Two leftover docs nits:

1. Please also update the hand-maintained API pages, not just the JSDoc. The published copy still says it “returns the value as specified in the `fallbackLocale` props”:
   - `docs/api/v11/general.md`
   - `docs/jp/api/v11/general.md`
   - `docs/zh/api/v11/general.md`

2. The JSDoc still has that same first sentence, then the new map-keys sentence right after it, which reads as contradictory. Please fold them into one description: string/array use the specified value, a map contributes its keys, and `default` is skipped.

The PR title/body are still about reading map values — worth updating those too so they match the current change.

haoku123:
Thanks for the review @kazupon! All three points addressed:

1. **Hand-maintained API docs** — updated `docs/api/v11/general.md`, `docs/jp/api/v11/general.md`, and `docs/zh/api/v11/general.md` with the map-keys description (each in its own language).
2. **JSDoc** — folded into a single sentence: string/array use the specified value, a map contributes its keys, and `default` is skipped.
3. **PR title/body** — retitled to 'fix(core-base): fallbackWithSimple uses map keys and skips default' and the description now reflects the current change (no longer mentions reading map values).

kazupon (approved):
Thanks!

PR #2571, opened 2026-08-17
fix(shared): avoid retaining array message references

Togetic:
### Description

Backport of #2554 to the `v11` line, as suggested in #2560.

`packages/shared/src/messages.ts` on `v11` is identical to the pre-#2554 `master` implementation (modulo one `eslint-disable` comment), so this is a straight cherry-pick of 0423e76 with your authorship preserved.

### One test omitted

`#2554` also added `returns an isolated copy of array messages` to `packages/core-base/test/context.test.ts`. That test uses `getLocaleMessage` as a standalone `core-base` export, which exists only on `master` (`packages/core-base/src/context.ts`) — `v11` has no such export, and no `core-base` test references it. On this line the equivalent surface is the Composer method, which is covered by `#2554`'s composer test, included here.

Everything else applied without conflict, so the backport carries three of the four tests:

- `packages/shared` — `deepCopy replaces arrays without retaining source references`
- `packages/shared` — `CVE-2024-52810 > __proto__ nested in array`
- `packages/vue-i18n-core` — `does not retain array references when merging messages`

### Verification

- `pnpm test:unit` (with `--typecheck`): 65 files, 885 passed / 2 skipped / 33 todo, no type errors.
- Reverting **only** `packages/shared/src/messages.ts` back to `v11`'s version fails exactly those three tests and nothing else (92 pass). So they are genuine regression tests on this line, not assertions that happen to hold either way.
- `eslint` and `prettier --list-different` clean on all three files.

### What the `__proto__` test shows on this line

Worth spelling out, since it reads like a CVE-2024-52810 bypass and isn't quite one. Because arrays were installed by reference, `deepCopy` never walked into them, so the `__proto__` key-skip never ran for anything nested inside an array: `{"list":[{"__proto__":{...}}]}` was copied through verbatim into stored messages. A `JSON.parse`d `__proto__` is an own property rather than a prototype write, so nothing is polluted by the copy itself — but the guard's intent stopped at the array boundary, and after this change it doesn't.

### Scope

`#2561` is deliberately **not** included — it is a shape-only refactor with no behavioural change, which I don't think earns its churn on a stable line. Happy to open it separately if you'd prefer the two lines to converge on the same implementation.

### DIFF
--- a/packages/shared/src/messages.ts
+++ b/packages/shared/src/messages.ts
@@ -17,20 +17,21 @@ export function deepCopy(src: any, des: any): void {
       if (key === '__proto__') {
         return
       }
-      // if src[key] is an object/array, set des[key]
-      // to empty object/array to prevent setting by reference
-      if (isObject(src[key]) && !isObject(des[key])) {
-        des[key] = Array.isArray(src[key]) ? [] : create()
-      }
 
-      if (isNotObjectOrIsArray(des[key]) || isNotObjectOrIsArray(src[key])) {
-        // replace with src[key] when:
-        // src[key] or des[key] is not an object, or
-        // src[key] or des[key] is an array
-        des[key] = src[key]
+      const value = src[key]
+      if (isArray(value)) {
+        // replace arrays instead of merging them, without retaining source references
+        const copied: unknown[] = []
+        copied.length = value.length
+        des[key] = copied
+        stack.push({ src: value, des: copied })
+      } else if (isObject(value)) {
+        if (!isObject(des[key]) || isArray(des[key])) {
+          des[key] = create()
+        }
+        stack.push({ src: value, des: des[key] })
       } else {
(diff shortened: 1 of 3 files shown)


Togetic:
@kazupon — following up on your "You can back-port to v11 branch! :)" in #2560. This is that backport, open since Aug 17.

It is a straight cherry-pick of `0423e76` from #2561 with your authorship preserved, and needed no adaptation: `v11`'s `packages/shared/src/messages.ts` matches the pre-#2554 master implementation apart from one eslint-disable comment. It still applies cleanly on `v11` as of today — this morning's three merges (#2577, #2578, #2580) don't touch that file.

One note in case the check list looks thin: CodeRabbit reports "review skipped" only because auto-reviews are disabled for base branches other than the default one, not because anything failed.

#2560 can close with this.

kazupon (approved):
Thanks!
LGTM!

PR #2574, opened 2026-08-18
fix(vue-i18n-core): tree-shake @vue/runtime-vapor from getCurrentInstance

### LINKED ISSUE #2573: `in` probe on the `vue` namespace defeats tree-shaking and pulls `@vue/runtime-vapor` into the bundle (Vue 3.6)
### Reporting a bug?

Since Vue 3.6, every app that bundles `vue-i18n` also bundles the entire Vapor runtime
(`@vue/runtime-vapor`, 306 756 bytes raw), even when the app never uses Vapor mode.

**Root cause**

1. The published ESM bundles open with a namespace import — `dist/vue-i18n.mjs:8`, and the same
   line in `vue-i18n.runtime.mjs`, `vue-i18n.node.mjs`, `vue-i18n.runtime.node.mjs`,
   `vue-i18n.esm-browser.js`, `vue-i18n.runtime.esm-browser.js`:

   ```js
   import * as Vue from 'vue'
   ```

2. Exactly one member of that namespace is ever consumed, through this helper
   (`dist/vue-i18n.mjs:280-289`):

   ```js
   function getCurrentInstance() {
       // NOTE(kazupon): avoid bundler warning
       const key = 'currentInstance';
       if (key in Vue) {
           return Vue[key];
       }
       else {
           return Vue.getCurrentInstance();
       }
   }
   ```

3. The namespace import on its own is harmless: bundlers happily prune `Vue.foo` when the access
   is static. The `in` probe is not. `key in Vue` forces the bundler to materialise the namespace
   object, which pins *every* export of `vue`.

4. Up to Vue 3.5 that only meant a slightly larger Vue runtime. Vue 3.6's bundler entry
   (`vue/dist/vue.runtime.esm-bundler.js:10`) now re-exports the Vapor runtime:

   ```js
   export * from "@vue/runtime-vapor";
   ```

   so the pinned namespace drags all of `@vue/runtime-vapor` in with it.

**Cost, measured on a real production build** (Vite + rolldown, 106 JS chunks, sizes through
`gzip -9`, identical method on both sides):

| Build | Total JS, gzip | Chunk that imports `vue-i18n` |
| --- | --- | --- |
| `vue-i18n` as published | 469 304 B | 386 039 B raw / 130 449 B gzip |
| `in` probe removed | 415 417 B | 251 110 B raw / 84 480 B gzip |
| **Difference** | **53 887 B gzip (+13.0 %)** | 134 929 B raw / 45 969 B gzip |

**Which part of the helper is responsible.** I patched the dependency at build time (Vite
`transform` hook) to isolate each factor. Same app, same Vue version, three builds:

| Helper body | Total JS, gzip | Vapor in the bundle |
| --- | --- | --- |
| as published (`const key = …; if (key in Vue)`) | 469 304 B | yes |
| `if ('currentInstance' in Vue) { return Vue['currentInstance'] } else { return Vue.getCurrentInstance() }` — literals inlined, probe kept | 469 322 B | yes |
| `return Vue.getCurrentInstance()` — probe removed, **namespace import left untouched** | 415 417 B | no |

So neither the namespace import nor the computed key is the blocker: **the `in` operator alone
pins the namespace**, and inlining the string literal does not help.

**The probe is unreachable anyway.** `vue` does not export `currentInstance` —
`@vue/runtime-core` 3.6.0-rc.3 contains no `as currentInstance` export — so `key in Vue` is always
`false` and `Vue.getCurrentInstance()` is always the branch taken. Same on 3.5.x.

### Expected behavior

`vue-i18n` should retain only `getCurrentInstance` from `vue`, so that the parts of the Vue runtime
an app does not use — Vapor above all — stay tree-shakeable.

Dropping the probe is sufficient, and behaviour-preserving since the branch it guards has never
been reachable:

```js
function getCurrentInstance() {
  return Vue.getCurrentInstance()
}
```

`import * as Vue from 'vue'` can stay exactly as it is, so the concern behind
`// NOTE(kazupon): avoid bundler warning` should be unaffected: the 415 417 B build above was
produced with the namespace import untouched, and no bundler warning was emitted.

If the probe is there to anticipate a future `currentInstance` export from `vue`, it would have to
be expressed without `in` — but a static `Vue.currentInstance` access on a missing export is
precisely what raises the bundler warning the comment refers to, so removing the branch until
`vue` actually ships that export looks like the cleaner trade-off.

### Reproduction

1. `npm i vue@3.6.0-rc.3 vue-i18n@11.4.8`
2. Build, in production mode with Vite (rolldown or Rollup), any app that imports `createI18n`
   from `vue-i18n` and contains no Vapor component at all.
3. Inspect the chunk holding `vue-i18n`: it carries `@vue/runtime-vapor` code
   (`createVaporApp`, `VaporFragment`, …).
4. Edit `node_modules/vue-i18n/dist/vue-i18n.mjs` so `getCurrentInstance()` reads
   `return Vue.getCurrentInstance();`, then rebuild: the Vapor code is gone and the bundle drops by
   the amounts in the table above.

Happy to put together a standalone repro repository if that helps — the one-line patch above is
the whole delta.

### System Info

```shell
vue:      3.6.0-rc.3
vue-i18n: 11.4.8
bundler:  Vite (rolldown), production build
node:     22
OS:       macOS host, Linux (Docker) build
```

### Screenshot

_No response_

### Additional context

A workaround, in case it is useful to other users while this is being looked at — a Vite plugin
that strips the probe from the published bundles (this is the exact "probe removed" variant
measured above):

```ts
import type { Plugin } from 'vite'

const VUE_I18N_MODULE = /[\\/]vue-i18n[\\/]dist[\\/]/
const PROBE =
  /const key = 'currentInstance';\s*if \(key in Vue\) \{\s*return Vue\[key\];\s*\}\s*else \{\s*return Vue\.getCurrentInstance\(\);\s*\}/

export function fixVueI18nTreeShaking(): Plugin {
  return {
    name: 'fix-vue-i18n-tree-shaking',
    enforce: 'pre',
    transform(code, id) {
      if (!VUE_I18N_MODULE.test(id) || !PROBE.test(code)) return
      return { code: code.replace(PROBE, 'return Vue.getCurrentInstance();'), map: null }
    }
  }
}
```

Related but different: #2234 reports a runtime error when `useI18n()` is called from a
`<script setup vapor>` component. This report is only about bundle size in apps that do **not**
use Vapor.

### Validations

lazerg:
## Summary

`getCurrentInstance` checks `'currentInstance' in Vue` before falling back to `Vue.getCurrentInstance()`. `currentInstance` is a real export on Vue 3.6's bundler build (it's the raw instance, unlike `getCurrentInstance()` which deliberately hides Vapor instances), so this isn't dead code, it's how Vapor component detection works here. The bug is narrower: the `in` operator itself is what pins the bundle. Testing a property with `in` forces bundlers to keep the whole `vue` namespace object alive, and since 3.6 re-exports `@vue/runtime-vapor` from that namespace, apps that never touch Vapor still end up with it in their bundle.

This swaps the `in` check for a static property read (`Vue.currentInstance`), which bundlers can prune when unused. Rebuilding `vue-i18n-core` and bundling a minimal `createI18n()` call with Rollup against `vue@3.6.0-rc.1` confirms it: `createVaporApp` and `VaporFragment` no longer show up in the output, and `pnpm test:vapor` still passes, so Vapor detection is unaffected.

One trade-off worth flagging: static access means a bundler that statically validates exports (Rollup does) will print a "not exported" warning when building against Vue < 3.6, since that export doesn't exist there. It's a build-time warning, not a runtime error, and bundle size on those versions doesn't change either way since `@vue/runtime-vapor` didn't exist before 3.6. Happy to adjust if there's a preference for silencing it instead.

Fixes #2573

### DIFF
--- a/packages/vue-i18n-core/src/utils.ts
+++ b/packages/vue-i18n-core/src/utils.ts
@@ -223,13 +223,12 @@ export function getCurrentInstance():
   | GenericComponentInstance
   | ComponentInternalInstance
   | null {
-  // NOTE(kazupon): avoid bundler warning
-  const key = 'currentInstance'
-  if (key in Vue) {
-    return (Vue as any)[key] as GenericComponentInstance | null
-  } else {
-    return Vue.getCurrentInstance()
-  }
+  // NOTE: static access (not `in`) so bundlers can tree-shake `@vue/runtime-vapor` when unused;
+  // this triggers a harmless "not exported" warning on Vue < 3.6, which doesn't have this export
+  const instance = (Vue as any).currentInstance
+  return instance !== undefined
+    ? (instance as GenericComponentInstance | null)
+    : Vue.getCurrentInstance()
 }
 
 /* eslint-enable @typescript-eslint/no-explicit-any */


kazupon:
@lazerg 
Thank you for your contributing!
Sorry, late my reply. 🙇 

I've already resolved another approach and released it.

I truly appreciate your contribution.
And thank you for using Vue I18n!

Issue #2621, opened 2026-09-06
inheritLocale setter does not write the inherited locale / fallbackLocale back to the core context


kazupon:
## Summary

`set inheritLocale` updates the Composer's refs but never assigns them to the core context:

```ts
// packages/vue-i18n-core/src/composer.ts:2501-2508
set inheritLocale(val: boolean) {
  _inheritLocale = val
  if (val && __root) {
    _locale.value = __root.locale.value as Locale
    _fallbackLocale.value = __root.fallbackLocale.value
    // _context.locale / _context.fallbackLocale are never assigned
    updateFallbackLocale(_context, _locale.value, _fallbackLocale.value)
  }
},
```

The two watchers immediately above it *do* assign them:

```ts
// packages/vue-i18n-core/src/composer.ts:2476-2489
watch(__root.locale, (val: Locale) => {
  if (_inheritLocale) {
    _locale.value = val
    _context.locale = val          // <-- present here
    updateFallbackLocale(_context, _locale.value, _fallbackLocale.value)
  }
})
watch(__root.fallbackLocale, (val: FallbackLocale) => {
  if (_inheritLocale) {
    _fallbackLocale.value = val
    _context.fallbackLocale = val  // <-- present here
    updateFallbackLocale(_context, _locale.value, _fallbackLocale.value)
  }
})
```

Across the whole file, `_context.locale` is assigned only at lines 2107 and 2479, and `_context.fallbackLocale` only at 2116 and 2486. The `inheritLocale` setter is the gap.

## Why this matters

The Composer has two independent readers of the active locale:

| reader | reads from |
| --- | --- |
| `t()`, `d()`, `n()` — via `translate()` / `resolveFormatLocale()` | `_context.locale` / `_context.fallbackLocale` |
| `te()`, `resolveMessages()`, `composer.locale.value`, `composer.fallbackLocale.value` | `_locale.value` / `_fallbackLocale.value` |

After `composer.inheritLocale = true` the two disagree: the Composer reports the inherited locale, while `t()` keeps translating in the one it had before.

`updateFallbackLocale()` on the last line does clear and re-warm the locale chain cache, but it warms it for the *new* values while `t()` goes on reading the *old* ones from the context, so it does not mask the problem.

The existing test only asserts the refs, which is why this was never caught:

```ts
// packages/vue-i18n-core/test/composer.test.ts:151-155
composer.inheritLocale = true
await nextTick()

expect(composer.locale.value).toEqual('en')                  // passes
[...]

Issue #2620, opened 2026-09-06
__localeChainCache grows without bound when fallbackLocale changes at runtime


kazupon:
## Summary

#2564 changed the locale chain cache from one dimension to two:

```diff
- __localeChainCache?: Map<Locale, Locale[]>
+ __localeChainCache?: Map<Locale, Map<string, Locale[]>>
```

The inner map is keyed by `friendlyJSONstringify(fallback)`. Entries are only ever added:

```ts
// packages/core-base/src/fallbacker.ts
chains.set(fallbackKey, chain)
```

There is no eviction of any kind — no `delete`, no size cap, no LRU, no `WeakRef`.

## Why this matters

Before #2564 the cache was bounded by the number of distinct start locales. It is now bounded by

```
number of start locales  x  number of distinct fallbackLocale serializations
```

`fallbackLocale` is not a constant in Vue I18n. It is exposed as a writable reactive value:

```ts
i18n.global.fallbackLocale.value = /* ... */
```

Any application that varies it — per user, per route, per tenant, or derived from request data — accumulates a permanent cache entry for every distinct value it has ever used. The entries are reachable from the core context for the whole lifetime of the i18n instance, so they are never collected.

## Reproduction

Save the script below and run it **from the repository root**:

```sh
npx tsx repro.ts
```

Observed with 10,000 distinct `fallbackLocale` values against a single start locale:

```
distinct fallbackLocale values used: 10,000
cache entries for start locale "en": 10,000
approximate retained size: ~1089 KB
eviction mechanism: none (no delete, no size cap, no LRU, no WeakRef)
```

The count grows linearly and nothing is ever released.

<details>
<summary>repro.ts</summary>

```ts
/**
 * Reproduction for issue-2: `__localeChainCache` grows without bound when
 * `fallbackLocale` varies at runtime.
 *
 * Usage (from the repository root):
 *   npx tsx repro.ts   # from the repository root
 */
import path from 'node:path'

const DISTINCT_FALLBACKS = 10_000
const START_LOCALE = 'en'

const rootDir = process.cwd() // run this from the repository root

// Build-time constants that the sources expect to be replaced by the bundler.
Object.assign(globalThis, {
  __DEV__: false,
  __TEST__: false,
  __VERSION__: '0.0.0',
  __BROWSER__: false,
  __GLOBAL__: false,
  __RUNTIME__: false,
  __BUNDLE_FILENAME__: 'repro',
  __ESM_BUNDLER__: true,
[...]

Issue #2619, opened 2026-09-06
fallbackWithLocaleChain rebuilds its cache key with JSON.stringify on every t() call


kazupon:
## Summary

`fallbackWithLocaleChain()` derives its cache key with `friendlyJSONstringify(fallback)`:

```ts
// packages/core-base/src/fallbacker.ts
const fallbackKey = friendlyJSONstringify(fallback)
let chains = context.__localeChainCache.get(startLocale)
// ...
const cached = chains.get(fallbackKey)
if (cached) {
  return cached
}
```

`friendlyJSONstringify` is `JSON.stringify()` plus three regex `String.prototype.replace()` passes:

```ts
// packages/shared/src/utils.ts
export const friendlyJSONstringify = (json: unknown): string =>
  JSON.stringify(json)
    .replace(/\u2028/g, '\\u2028')
    .replace(/\u2029/g, '\\u2029')
    .replace(/\u0027/g, '\\u0027')
```

The key has to be built *before* the map lookup can happen, so this cost is paid **on every call, including cache hits**.

## Why this is on the hot path

```
translate()                      packages/core-base/src/translate.ts:613
  -> resolveMessageFormat()      packages/core-base/src/translate.ts:733
    -> localeFallbacker()        packages/core-base/src/translate.ts:742
      -> fallbackWithLocaleChain()
```

`translate()` calls `resolveMessageFormat()` for every non-`resolvedMessage` translation, so every `t()` call now serializes the whole `fallbackLocale` value.

The cache exists to avoid recomputing the chain, but the lookup itself became far more expensive than the lookup it replaced.

## Measurements

Real `fallbackWithLocaleChain` from source, cache-hit path, N = 1,000,000, Node 24.20.0:

| `fallbackLocale` shape | before (`1dfed77c^`) | after (`master`) | ratio | added per call |
| --- | ---: | ---: | ---: | ---: |
| string (`'en'`) | 20 ms | 138 ms | 6.8x | +118 ns |
| array (`['en','ja']`) | 24 ms | 166 ms | 6.8x | +142 ns |
| map, 5 locales | 24 ms | 334 ms | 13.8x | +309 ns |
| map, 30 locales | 24 ms | 1788 ms | **74.5x** | **+1764 ns** |

Numbers vary by a few percent between runs; the ratios are stable.

An app with a 30-locale `fallbackLocale` map pays roughly +1.8 microseconds per `t()` call.

## Reproduction

Save the script below and run it **from the repository root**:

[...]

Issue #2600, opened 2026-08-25
rt() throws "Cannot convert object to primitive value" on re-render in 11.4.9


jurgensweere:
### Reporting a bug?

Since v11.4.9, calling `rt()` on items returned by `tm()` throws `TypeError: Cannot convert object to primitive value` as soon as the component re-renders. The first render is fine; every render after that throws.
This happens when the messages are precompiled by @intlify/unplugin-vue-i18n (SFC `<i18n>` block)
It works in 11.4.8 and breaks in 11.4.9.

I believe this is a regression from #2571 ("fix(shared): avoid retaining array message references"). That PR changed `deepCopy` so objects inside arrays are now recreated with `create()` (`Object.create(null)`) instead of being kept by reference. Precompiled messages are message AST objects, so every AST node inside an array loses its prototype. `rt(astNode)` passes that object on as the `key`, and the dev-only devtools timeline code in `compileMessageFormat` does `groupId: `${'translate'}:${key}``. Calling `String()` on a null-prototype object throws.

This also explains why only the second render fails: the devtools emitter is attached in a global mixin `mounted()` hook, so the first render has no emitter and never reaches that line.

### Expected behavior

`rt()` on items from `tm()` should keep working on every render, like it did in 11.4.8. Message AST nodes should not lose their prototype when messages are merged.

### Reproduction

https://stackblitz.com/edit/vitejs-vite-cdp1et9f

Changing language will trigger the console error

### System Info

```shell
System:
    OS: macOS 26.6.2
    CPU: (8) arm64 Apple M1 Pro
    Memory: 80.38 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.18.0 - /Users/jurgensweere/.nvm/versions/node/v24.18.0/bin/node
    npm: 11.18.0 - /Users/jurgensweere/.nvm/versions/node/v24.18.0/bin/npm
    pnpm: 11.15.1 - /Users/jurgensweere/.nvm/versions/node/v24.18.0/bin/pnpm
  Browsers:
    Chrome: 151.0.7922.174
    Edge: 151.0.4129.101
    Firefox: 137.0.2
    Safari: 26.6.2
  npmPackages:
    @intlify/unplugin-vue-i18n: ^11.2.5 => 11.2.5 
    @vitejs/plugin-vue: ^6.0.8 => 6.0.8 
    @vue/language-core: ^3.3.11 => 3.3.11 
[...]

PR #2622, opened 2026-09-06
fix(core-base): cache locale chains by fallback identity (#2620 #2619 #2621)

haoku123:
## Summary

This PR fixes three related issues in the locale-chain cache:

- **#2620** — `__localeChainCache` grows without bound when `fallbackLocale` changes at runtime.
- **#2619** — `fallbackWithLocaleChain` rebuilt a `JSON.stringify`-derived cache key on every `t()` call.
- **#2621** — the `Composer.inheritLocale` setter copied the inherited locale/fallback into the reactive refs but never wrote them back to the core context.

## Root cause & fix

`fallbackWithLocaleChain` keyed its cache by `friendlyJSONstringify(fallback)` inside a per-start-locale `Map`. When the fallback value's identity changed (a freshly created array, e.g. from a reactive computed, or repeated `i18n.global.fallbackLocale.value = …`), every distinct serialized value became a new key that was never evicted → unbounded growth (#2620). The `JSON.stringify` was also recomputed on every call even when the cached value was returned (#2619).

Replaced the string-keyed `Map` with a new `LocaleChainCache` that keys by fallback **identity**:
- object/array fallbacks → `WeakMap` keyed by the value itself, so entries are released together with the fallback object and never accumulate;
- primitive fallbacks (`string | false`) → `Map` keyed by the value itself, removing the per-call `JSON.stringify`.

This keeps the root context's cache slot (queried by `getMessageContextOptions` with a local composer's fallback) logically separate from the local composer's slot — no whole-cache invalidation, so no thrash. In-place mutation of a fallback object is still handled through the existing `updateFallbackLocale()` hook (which now drops the whole cache, recreating it lazily).

For #2621: the `inheritLocale` setter now also writes `_context.locale` / `_context.fallbackLocale`, matching the `fallbackLocale` setter and the root watcher.

## Test plan

- Updated the pre-existing `fallbacker` cache-invalidation test: in-place mutation now requires `updateFallbackLocale` (the documented contract for identity keying).
- Added a regression test asserting the cache is keyed by fallback identity so distinct object fallbacks do not collide (#2619).
- Added a regression test for the `inheritLocale` write-back (#2621).
- `core-base` fallbacker suite (46) and composer `inheritLocale` suite (4) pass.

Note: the repo's `datetime.test.ts` / composer `parts formatting` tests fail on `master` too in this environment (timezone-dependent), so they are unrelated to this change.

I implemented the identity-keying approach suggested in #2619 (WeakMap for object/array fallbacks, plain Map for primitives) rather than whole-cache invalidation, since that would thrash when the root context is queried with a local composer's fallback in alternation.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved locale fallback handling by excluding the `default` entry where appropriate.
  - Updated locale inheritance so locale and fallback settings stay synchronized when inheritance is enabled.
  - Improved resolution of merged array-based messages.
  - Enhanced fallback cache handling for changing fallback configurations.

- **Developer Tools**
  - Standardized grouping for missing-message and fallback events in devtools.

### DIFF
--- a/packages/core-base/src/context.ts
+++ b/packages/core-base/src/context.ts
@@ -14,6 +14,7 @@ import {
   warnOnce
 } from '@intlify/shared'
 import { fallbackWithSimple } from './fallbacker'
+import type { LocaleChainCache } from './fallbacker'
 import { resolveWithKeyValue } from './resolver'
 import { CoreWarnCodes, getWarnMessage } from './warnings'
 
@@ -310,7 +311,7 @@ export type CoreContext<
 export interface CoreInternalContext {
   __datetimeFormatters: Map<string, Intl.DateTimeFormat>
   __numberFormatters: Map<string, Intl.NumberFormat>
-  __localeChainCache?: Map<Locale, Map<string, Locale[]>>
+  __localeChainCache?: LocaleChainCache
   __v_emitter?: VueDevToolsEmitter
 }
 
@@ -657,7 +658,10 @@ export function updateFallbackLocale<Message = string>(
   fallback: FallbackLocale
 ): void {
   const context = ctx as unknown as CoreInternalContext
-  context.__localeChainCache = new Map()
+  // Drop the whole cache: it is recreated lazily on the next fallback lookup.
+  // This also handles in-place mutations of the fallback value, which a WeakMap
+  // (used for object/array fallbacks) cannot invalidate on its own.
+  context.__localeChainCache = undefined
   ctx.localeFallbacker<Message>(ctx, fallback, locale)
 }
 
--- a/packages/core-base/src/fallbacker.ts
+++ b/packages/core-base/src/fallbacker.ts
@@ -1,5 +1,4 @@
 import {
-  friendlyJSONstringify,
   isArray,
   isBoolean,
   isFunction,
@@ -119,6 +118,48 @@ export function fallbackWithSimple<Message = string>(
  *
  * @VueI18nGeneral
  */
+/**
+ * Cache of locale chains keyed by the fallback locale *identity* rather than a
+ * `JSON.stringify`-derived string (see #2619 / #2620).
+ *
+ * - Object/array fallbacks are stored in a `WeakMap` keyed by the fallback
+ *   value itself. Entries are released together with the fallback object, so
+ *   repeatedly creating new fallback arrays no longer grows the cache without
+ *   bound (#2620).
+ * - Primitive fallbacks (`string | false`) are stored in a plain `Map` keyed by
+ *   the value itself, which avoids rebuilding a string key on every `t()` call
+ *   (#2619).
+ */
+class LocaleChainCache {
+  private _weak = new WeakMap<object, Map<Locale, Locale[]>>()
+  private _primitive = new Map<string | false, Map<Locale, Locale[]>>()
+
+  get(fallback: FallbackLocale, startLocale: Locale): Locale[] | undefined {
+    if (isObject(fallback)) {
+      return this._weak.get(fallback as object)?.get(startLocale)
+    }
+    return this._primitive.get(fallback as string | false)?.get(startLocale)
+  }
+
+  set(fallback: FallbackLocale, startLocale: Locale, chain: Locale[]): void {
+    if (isObject(fallback)) {
+      let inner = this._weak.get(fallback as object)
+      if (!inner) {
+        inner = new Map<Locale, Locale[]>()
+        this._weak.set(fallback as object, inner)
+      }
+      inner.set(startLocale, chain)
+    } else {
+      let inner = this._primitive.get(fallback as string | false)
+      if (!inner) {
+        inner = new Map<Locale, Locale[]>()
+        this._primitive.set(fallback as string | false, inner)
+      }
+      inner.set(startLocale, chain)
+    }
+  }
+}
+
 export function fallbackWithLocaleChain<Message = string>(
   ctx: CoreContext<Message>,
   fallback: FallbackLocale,
@@ -128,17 +169,10 @@ export function fallbackWithLocaleChain<Message = string>(
   const context = ctx as unknown as CoreInternalContext
 
   if (!context.__localeChainCache) {
-    context.__localeChainCache = new Map()
+    context.__localeChainCache = new LocaleChainCache()
   }
 
-  const fallbackKey = friendlyJSONstringify(fallback)
-  let chains = context.__localeChainCache.get(startLocale)
-  if (!chains) {
-    chains = new Map()
-    context.__localeChainCache.set(startLocale, chains)
-  }
-
-  const cached = chains.get(fallbackKey)
+  const cached = context.__localeChainCache.get(fallback, startLocale)
   if (cached) {
     return cached
   }
@@ -166,7 +200,7 @@ export function fallbackWithLocaleChain<Message = string>(
   if (isArray(block)) {
     appendBlockToChain(chain, block, false)
   }
-  chains.set(fallbackKey, chain)
+  context.__localeChainCache.set(fallback, startLocale, chain)
 
   return chain
 }
--- a/packages/core-base/test/fallbacker.test.ts
+++ b/packages/core-base/test/fallbacker.test.ts
@@ -1,4 +1,4 @@
-import { createCoreContext as context } from '../src/context'
+import { createCoreContext as context, updateFallbackLocale } from '../src/context'
 import { CoreErrorCodes, errorMessages } from '../src/errors'
 import { fallbackWithLocaleChain, fallbackWithSimple, resolveLocale } from '../src/fallbacker'
 
@@ -277,12 +277,27 @@ describe('fallbackWithLocaleChain', () => {
       ])
     })
 
-    test('recomputes the chain when the fallback is mutated in place', () => {
+    test('recomputes the chain when the fallback is mutated in place (via updateFallbackLocale)', () => {
       const fallback: Record<string, string[]> = { sl: ['en'], default: ['en'] }
       expect(fallbackWithLocaleChain(ctx, fallback, 'sl')).toEqual(['sl', 'en'])
 
       fallback.sl = ['de']
+      // Mutation of a fallback object is detected through the explicit
+      // invalidation hook, since the cache is now keyed by fallback identity
+      // rather than a serialized key (#2619 / #2620).
+      updateFallbackLocale(ctx, 'sl', fallback)
       expect(fallbackWithLocaleChain(ctx, fallback, 'sl')).toEqual(['sl', 'de', 'en'])
+    })
+
+    test('keys the cache by fallback identity so distinct object fallbacks do not collide (#2619)', () => {
+      const a = ['en', 'ja']
+      const b = ['en', 'ja'] // structurally equal, different reference
+      const chainA = fallbackWithLocaleChain(ctx, a, 'fr')
+      const chainB = fallbackWithLocaleChain(ctx, b, 'fr')
+      expect(chainA).toEqual(chainB)
+      // each reference keeps its own cache slot
+      expect(fallbackWithLocaleChain(ctx, a, 'fr')).toBe(chainA)
+      expect(fallbackWithLocaleChain(ctx, b, 'fr')).toBe(chainB)
     })
 
     test('keeps cache hit when the same fallback is used again (#2562)', () => {
--- a/packages/vue-i18n-core/src/composer.ts
+++ b/packages/vue-i18n-core/src/composer.ts
@@ -2502,7 +2502,9 @@ export function createComposer(options: any = {}): ComposerInternalInstance {
       _inheritLocale = val
       if (val && __root) {
         _locale.value = __root.locale.value as Locale
+        _context.locale = _locale.value
         _fallbackLocale.value = __root.fallbackLocale.value
+        _context.fallbackLocale = _fallbackLocale.value
         updateFallbackLocale(_context, _locale.value, _fallbackLocale.value)
       }
     },
--- a/packages/vue-i18n-core/test/composer.test.ts
+++ b/packages/vue-i18n-core/test/composer.test.ts
@@ -19,7 +19,8 @@ import {
   DatetimePartsSymbol,
   EnableEmitter,
   NumberPartsSymbol,
-  TranslateVNodeSymbol
+  TranslateVNodeSymbol,
+  CoreContextSymbol
 } from '../src/symbols'
 import { getWarnMessage, I18nWarnCodes } from '../src/warnings'
 
@@ -153,6 +154,26 @@ describe('inheritLocale', () => {
 
     expect(composer.locale.value).toEqual('en')
     expect(composer.fallbackLocale.value).toEqual(['ja', 'fr'])
+  })
+
+  test('writes inherited locale/fallbackLocale back to the core context (#2621)', () => {
+    const root = createComposer({
+      locale: 'en',
+      fallbackLocale: ['ja', 'fr']
+    })
+    const composer = createComposer({
+      locale: 'ja',
+      fallbackLocale: ['zh', 'de'],
+      inheritLocale: false,
+      __root: root
+    })
+    const ctx = (composer as unknown as Record<symbol, any>)[CoreContextSymbol]
+    expect(ctx.locale).toEqual('ja')
+    expect(ctx.fallbackLocale).toEqual(['zh', 'de'])
+
+    composer.inheritLocale = true
+    expect(ctx.locale).toEqual('en')
+    expect(ctx.fallbackLocale).toEqual(['ja', 'fr'])
   })
 })
 


packages/core-base/src/fallbacker.ts
```diff
@@ -119,6 +118,48 @@ export function fallbackWithSimple<Message = string>(
...
+ *
+ * - Object/array fallbacks are stored in a `WeakMap` keyed by the fallback
+ *   value itself. Entries are released together with the fallback object, so
+ *   repeatedly creating new fallback arrays no longer grows the cache without
+ *   bound (#2620).
+ * - Primitive fallbacks (`string | false`) are stored in a plain `Map` keyed by
+ *   the value itself, which avoids rebuilding a string key on every `t()` call
+ *   (#2619).
+ */
+class LocaleChainCache {
```
kazupon, line 133: We don't use `class` in the implementation — bundle size matters a lot here, and there is no
class under `packages/*/src` today. This one would be the first :)

It also doesn't build as it stands, because the class isn't exported. The CI Build job runs
`pnpm build --all -t`, and the type build fails there:

```
[MISSING_EXPORT] "LocaleChainCache" is not exported by
"temp/packages/core-base/src/fallbacker.d.ts"
```

`pnpm typecheck` reports the same thing:

```
packages/core-base/src/context.ts(17,15): error TS2459:
Module '"./fallbacker"' declares 'LocaleChainCache' locally, but it is not exported.
```

Exporting it isn't quite what we want either, since `CoreInternalContext` goes into
`core-base.d.ts` and the class would become a public type.

Could you inline it into the function instead? 🙏

```ts
export function fallbackWithLocaleChain<Message = string>(
  ctx: CoreContext<Message>,
  fallback: FallbackLocale,
  start: Locale
): Locale[] {
  const startLocale = isString(start) ? start : DEFAULT_LOCALE
  const context = ctx as unknown as CoreInternalContext

  // NOTE: chains are cached per fallback value, not per serialized content.
  // object / array fallbacks are keyed by identity in a WeakMap so that their
  // entries are released together with the fallback itself.
  let chains: Map<Locale, Locale[]> | undefined
  if (fallback === context.__lastFallback && context.__lastChains) {
    chains = context.__lastChains
  } else {
    if (isObject(fallback)) {
      if (!context.__localeChainObjectCache) {
        context.__localeChainObjectCache = new WeakMap()
      }
      chains = context.__localeChainObjectCache.get(fallback)
      if (!chains) {
        chains = new Map()
        context.__localeChainObjectCache.set(fallback, chains)
      }
    } else {
      if (!context.__localeChainCache) {
        context.__localeChainCache = new Map()
      }
      chains = context.__localeChainCache.get(fallback)
      if (!chains) {
        chains = new Map()
        context.__localeChainCache.set(fallback, chains)
      }
    }
    context.__lastFallback = fallback
    context.__lastChains = chains
  }

  const cached = chains.get(startLocale)
  if (cached) {
    return cached
  }

  // ... the chain building part stays as it is ...

  chains.set(startLocale, chain)
  return chain
}
```

`fallback` works as the Map key directly, so the `as` casts go away too. The one-slot memo at the
top is for performance — see my other comment.

packages/core-base/src/fallbacker.ts
```diff
@@ -119,6 +118,48 @@ export function fallbackWithSimple<Message = string>(
  *
  * @VueI18nGeneral
  */
+/**
```
kazupon, line 121: The class landed between the JSDoc of `fallbackWithLocaleChain` (104-120) and the function
itself (163), so `@VueI18nSee` and `@VueI18nGeneral` are now attached to the class instead of
the public function.

If you inline the cache as in the other comment this goes away by itself. Otherwise please move
the declaration above that JSDoc block. 🙏

packages/core-base/src/context.ts
```diff
@@ -9,10 +9,12 @@ import {
   isPlainObject,
   isRegExp,
   isString,
+  toDevtoolsGroupId,
   warn,
   warnOnce
 } from '@intlify/shared'
 import { fallbackWithSimple } from './fallbacker'
+import type { LocaleChainCache } from './fallbacker'
```
kazupon, line 17: Along with the above, please drop `import type { LocaleChainCache }` and use plain fields:

```ts
export interface CoreInternalContext {
  __datetimeFormatters: Map<string, Intl.DateTimeFormat>
  __numberFormatters: Map<string, Intl.NumberFormat>
  /** locale chains for a primitive `fallbackLocale`; bounded by the locale count */
  __localeChainCache?: Map<Locale | false | undefined, Map<Locale, Locale[]>>
  /** locale chains for an object / array `fallbackLocale`; released with the fallback */
  __localeChainObjectCache?: WeakMap<object, Map<Locale, Locale[]>>
  /** one-slot memo for the most recently used fallback */
  __lastFallback?: FallbackLocale
  __lastChains?: Map<Locale, Locale[]>
  __v_emitter?: VueDevToolsEmitter
}
```

Same style as `__datetimeFormatters` / `__numberFormatters`, and no new public type.

packages/core-base/src/context.ts
```diff
@@ -656,7 +658,10 @@ export function updateFallbackLocale<Message = string>(
   fallback: FallbackLocale
 ): void {
   const context = ctx as unknown as CoreInternalContext
-  context.__localeChainCache = new Map()
+  // Drop the whole cache: it is recreated lazily on the next fallback lookup.
+  // This also handles in-place mutations of the fallback value, which a WeakMap
+  // (used for object/array fallbacks) cannot invalidate on its own.
+  context.__localeChainCache = undefined
```
kazupon, line 664: Dropping the cache and recreating it lazily is fine, but we need to clear more than this context.

`getMessageContextOptions()` resolves linked messages against `fallbackContext` (the root) while
passing **this** composer's `fallbackLocale`, so the root caches a chain under the same fallback
identity. Clearing only the local context leaves that root entry stale, and a linked message
keeps returning the old chain.

This one is my miss in #2619, sorry 🙇 It reproduces like this — passes on master, fails with
this PR:

```ts
const root = createComposer({
  locale: 'sl',
  fallbackLocale: false,
  messages: { sl: {}, en: { msg: 'from en' }, de: { msg: 'from de' } },
  missingWarn: false,
  fallbackWarn: false
})
const local = createComposer({
  locale: 'sl',
  fallbackLocale: ['en'],
  inheritLocale: false,
  __root: root,
  messages: { sl: { linked: '@:msg' } }
})

expect(local.t('linked')).toBe('from en')
;(local.fallbackLocale.value as string[])[0] = 'de'
local.fallbackLocale.value = local.fallbackLocale.value
expect(local.t('linked')).toBe('from de')
```

Note the deep watcher in my other comment does **not** cover this — it invalidates the local
context only.

Could you clear both here? 🙏

```ts
const context = ctx as unknown as CoreInternalContext
context.__localeChainCache = undefined
context.__localeChainObjectCache = undefined
context.__lastFallback = undefined
context.__lastChains = undefined

// linked messages resolve against the root with *this* composer's fallback,
// so the root caches a chain under the same fallback identity.
const root = ctx.fallbackContext as unknown as CoreInternalContext | undefined
if (root) {
  root.__localeChainCache = undefined
  root.__localeChainObjectCache = undefined
  root.__lastFallback = undefined
  root.__lastChains = undefined
}

ctx.localeFallbacker<Message>(ctx, fallback, locale)
```

Clearing the root wholesale is a bit blunt, but `fallbackLocale` changes are rare so I think it's
fine. If you'd rather be precise, deleting just the entry for `fallback` from the root's two
caches (plus resetting its memo when it points at the same object) works too.

If we forget the two memo lines, the memo hands back a stale `chains` even after
`updateFallbackLocale()`.

Last thing: the comment says this also handles in-place mutations, but that's only true when
`updateFallbackLocale()` actually gets called. Right now the Composer calls it on re-assignment
only, so `i18n.global.fallbackLocale.value.ja = [...]` never reaches here. Either we add the
watcher (next comment) or we reword this.

packages/vue-i18n-core/src/composer.ts
```diff
@@ -2501,7 +2502,9 @@ export function createComposer(options: any = {}): ComposerInternalInstance {
       _inheritLocale = val
       if (val && __root) {
         _locale.value = __root.locale.value as Locale
+        _context.locale = _locale.value
         _fallbackLocale.value = __root.fallbackLocale.value
+        _context.fallbackLocale = _fallbackLocale.value
```
kazupon, line 2507: Could you add a deep watcher for `fallbackLocale`? 🙏 It belongs next to the
`watch root locale & fallbackLocale` block just above (around line 2474) — I'm commenting here
because that line isn't part of the diff.

```ts
// invalidate the locale chain cache when `fallbackLocale` is mutated in place.
// `flush: 'sync'` so that a `t()` called right after the mutation sees the new chain.
watch(
  _fallbackLocale,
  () => {
    updateFallbackLocale(_context, _locale.value, _context.fallbackLocale as FallbackLocale)
  },
  { deep: true, flush: 'sync' }
)
```

Please put it outside `if (__root && inBrowser)` — the root composer needs it too.

Without it, mutating the fallback in place stops updating translations. This passes on master and
fails with this PR:

```ts
const composer = createComposer({
  locale: 'sl',
  fallbackLocale: { sl: ['en'], default: ['en'] },
  messages: { sl: {}, en: { msg: 'from en' }, de: { msg: 'from de' } },
  missingWarn: false,
  fallbackWarn: false
})

expect(composer.t('msg')).toBe('from en')
;(composer.fallbackLocale.value as Record<string, string[]>).sl = ['de']
expect(composer.t('msg')).toBe('from de')
```

`await nextTick()` doesn't help, and `fallbackLocale: ['en']` with `.splice(0, 1, 'de')` behaves
the same way.

The reason I care: #2564 keyed by content, so in-place mutation just worked, and that shipped in
v11.4.9 / v11.4.10. Identity keying loses it, and I don't want to take that away in a patch
release when we back-port.

`flush: 'sync'` rather than `'pre'` because a `t()` right after the mutation has to see the new
chain, which is how it behaves today.

Note this only works in the browser, since `_ref = inBrowser ? ref : shallowRef`. In SSR Vue
itself doesn't track the mutation either, so we're consistent there :)

packages/core-base/src/fallbacker.ts
```diff
@@ -119,6 +118,48 @@ export function fallbackWithSimple<Message = string>(
...
+ *   bound (#2620).
+ * - Primitive fallbacks (`string | false`) are stored in a plain `Map` keyed by
+ *   the value itself, which avoids rebuilding a string key on every `t()` call
+ *   (#2619).
+ */
+class LocaleChainCache {
+  private _weak = new WeakMap<object, Map<Locale, Locale[]>>()
+  private _primitive = new Map<string | false, Map<Locale, Locale[]>>()
+
+  get(fallback: FallbackLocale, startLocale: Locale): Locale[] | undefined {
```
kazupon, line 137: Two map lookups alone don't get us back to where we were before #2564. I measured **+25ns per
call**, about 2.1x the old baseline. It's constant across shapes now, which is the important
part, and it's a big win over current master either way.

With the one-slot memo it's **+2-3ns**, basically the same as before #2564.

Cache-hit path, N=1,000,000, Node 24.20.0:

| `fallbackLocale` | before #2564 | current master | 2 lookups | + memo |
| --- | ---: | ---: | ---: | ---: |
| string (`'en'`) | 20 ms | 138 ms | 46 ms | **23 ms** |
| array | 24 ms | 166 ms | 47 ms | **25 ms** |
| map, 5 locales | 24 ms | 334 ms | 47 ms | **25 ms** |
| map, 30 locales | 24 ms | 1788 ms | 48 ms | **25 ms** |

`t()` is a hot path, so I think it's worth it. The memo only misses when the root and a local
composer's `fallbackLocale` alternate, and then it just falls back to the two lookups.

packages/core-base/test/fallbacker.test.ts
```diff
@@ -267,14 +277,29 @@ describe('fallbackWithLocaleChain', () => {
...
 
       fallback.sl = ['de']
+      // Mutation of a fallback object is detected through the explicit
+      // invalidation hook, since the cache is now keyed by fallback identity
+      // rather than a serialized key (#2619 / #2620).
+      updateFallbackLocale(ctx, 'sl', fallback)
       expect(fallbackWithLocaleChain(ctx, fallback, 'sl')).toEqual(['sl', 'de', 'en'])
     })
 
+    test('keys the cache by fallback identity so distinct object fallbacks do not collide (#2619)', () => {
```
kazupon, line 292: This one covers identity keying, but we don't have a regression test for #2620 itself yet.
Could you add one that throws a lot of distinct object fallbacks at it and checks the strong
`Map` stays empty? 🙏

```ts
test('does not accumulate entries for distinct object fallbacks (#2620)', () => {
  for (let i = 0; i < 1000; i++) {
    fallbackWithLocaleChain(ctx, { ja: ['en'], default: [`fb-${i}`] }, 'en')
  }
  // object fallbacks live in the WeakMap; the strong Map must stay empty
  expect((ctx as unknown as CoreInternalContext).__localeChainCache).toBeUndefined()
})
```

We'll also want the two repros from my other comments as regression tests in `composer.test.ts`
— the in-place mutation one and the linked-message one. It runs in jsdom, so `_ref` is `ref` and
the mutation is observable. With `flush: 'sync'` there shouldn't be any need for
`await nextTick()`.

kazupon (changes requested):
Thank you for the PR! 🙏

The direction is exactly what I had in mind in #2619 — keying by fallback identity, `WeakMap`
for objects and `Map` for primitives. And thanks for picking up #2621 as well, I confirmed the
write-back fix works on the actual translation output, not just on the refs.

A few things before we can merge.

**1. Could you rebase onto master?**

This branch is based on `1dfed77c` (#2564 itself), so it's 42 commits behind. Because of that,
`toDevtoolsGroupId` is imported but it doesn't exist on this branch yet — we added it in #2605.
The build doesn't pass as it stands.

The same reason brings in changes that are already on master (the `default` filter in
`fallbackWithSimple` from #2580, and the `tm` test from #2600). They should disappear after the
rebase. Let's keep this PR to #2619 / #2620 / #2621 only.

**2. `LocaleChainCache` breaks the type build, and we don't use classes here.**

By the way, GitHub Actions hasn't run on this PR yet — the green check is CodeRabbit only. Let me
approve the workflow so we get real signal. When it runs, the Build job (`pnpm build --all -t`)
will fail, because the class isn't exported.

On top of that we care a lot about bundle size, so there is no `class` anywhere under
`packages/*/src` today. Could you inline it into plain fields on `CoreInternalContext`? 🙏
That fixes both at once. Details inline.

**3. In-place mutation of `fallbackLocale` stops working, in two different ways.**

This is the part I'd most like to get right before we back-port, because the behavior shipped in
v11.4.9 / v11.4.10 and users would silently break.

- The Composer needs to invalidate when the fallback object is mutated in place.
- `updateFallbackLocale()` also has to clear the **root** context, not only its own. Linked
  messages resolve against the root with the local composer's fallback, so the root keeps a
  stale chain even after the local one is dropped.

The second one is my miss in #2619 — sorry about that 🙇 Details inline.

I also left comments on performance and tests.

One small thing I couldn't attach inline because it's outside the diff: at `composer.ts:2116`,
`_context.fallbackLocale` (raw) and `_fallbackLocale.value` (a reactive proxy in the browser)
are different objects. With identity keying that means two cache entries for the same fallback,
since `t()` passes the former and `te()` the latter. They're in the `WeakMap` so they get
released and it isn't a correctness problem, but swapping the assignment order lines them up:

```ts
set: val => {
  _fallbackLocale.value = val
  _context.fallbackLocale = _fallbackLocale.value
  updateFallbackLocale(_context, _locale.value, _fallbackLocale.value)
}
```

Same for the root watcher at 2486. A separate PR is completely fine.

Thanks!

haoku123:
Rebased onto latest `master` (123d4076) — the merge conflict is resolved.

The three commits that landed upstream in the meantime only touched `package.json` / `pnpm-lock.yaml` (rolldown, oxc-parser), so none of the source files this PR modifies were affected: the rebase only moves the parent commit, the diff is unchanged (still 5 files, +92/-16).

Verified locally:
- `packages/core-base/test/fallbacker.test.ts` — passes.
- `packages/vue-i18n-core/test/composer.test.ts` — 96/96 pass with `TZ=UTC`.

Note: without a fixed `TZ` there is one timezone-dependent failure in the datetime "parts formatting" case (hour/dayPeriod). That one also fails on a clean checkout without this change, so it is unrelated to this PR.

kazupon:
Thanks for rebasing! I've confirmed that the merge conflict is resolved. 🙏

Could you also address the remaining comments from my review? In particular, we still need to replace the class-based cache implementation, handle in-place fallback mutations, invalidate the root cache used by linked messages, and add the corresponding regression tests.

CI is currently failing with a type error because `LocaleChainCache` is not exported. This is separate from the timezone-dependent test failure and also needs to be resolved.

The details are in the existing inline comments. Please feel free to ask in the relevant threads if anything is unclear!

kazupon:
@haoku123
Thank you for picking up all three issues so quickly, and for rebasing right away! 🙏
The cache is now keyed by fallback identity, the same direction as this PR, and your
`inheritLocale` write-back for #2621 was correct too.

I'm sorry for taking this over without asking you first 🙇
The regression had already shipped in v11.4.9 / v11.4.10, so I wanted to get a fix out soon:

- #2619 / #2620: #2633 (master), #2635 (v11)
- #2621: #2634 (master), #2636 (v11)

The v11 fixes are released in v11.4.12.

I also need to apologize for two of my review comments, which turned out to be wrong:

- The one-slot memo isn't worth it. In the production bundle, the two map lookups are
  only 2-3 ns slower, so #2633 doesn't have it.
- The root invalidation I suggested in `updateFallbackLocale()` doesn't work when it's
  called from the Composer's `fallbackLocale` setter, because `fallbackContext` only
  points at the root while a translation runs. #2633 handles this in the Composer instead.

Sorry for pointing you in the wrong direction there.

Since everything is fixed now, let me close this PR. Thanks again for all your contributions!

PR #2633, opened 2026-09-16
fix(core-base): cache locale chains by fallback identity

### LINKED ISSUE #2619: fallbackWithLocaleChain rebuilds its cache key with JSON.stringify on every t() call

### LINKED ISSUE #2620: __localeChainCache grows without bound when fallbackLocale changes at runtime

kazupon:
### Description

Fixes #2619
Fixes #2620

#2564 (shipped to v11 as #2577, in v11.4.9 / v11.4.10) keyed the locale chain cache by `friendlyJSONstringify(fallback)`. That was the fix for #2562, but it introduced two problems:

- every `t()` call serialized the whole `fallbackLocale` before it could look the chain up, including on cache hits (#2619)
- entries were only ever added, so the cache grew with every distinct `fallbackLocale` the app had ever used (#2620)

This keys the cache by the **identity** of the fallback instead:

- object / array fallbacks → `WeakMap`, released together with the fallback
- primitive fallbacks → `Map`, bounded by the number of locales

`updateFallbackLocale()` stays the explicit invalidation hook, and now drops only the entry for that fallback, from both the context and the root context.

### Why the root context is invalidated too

Linked-message resolution asks the **root** context with the **local** composer's `fallbackLocale` (`getMessageContextOptions()`), so the root caches a chain for that fallback as well. Clearing only the local context leaves a stale chain there.

The core context only points at the root while a translation runs (`wrapWithDeps`), so the Composer points it there while invalidating, too.

The whole root cache is not thrown away, because it is shared by every composer.

### In-place mutation

With an identity key, mutating a fallback object in place no longer invalidates the chain by itself. The Composer now watches `fallbackLocale` deeply (`flush: 'sync'`) and invalidates the chain when its contents change, so this keeps working as in v11.4.9 / v11.4.10.

This only works in the browser, where `fallbackLocale` is deeply reactive. On the server it is a `shallowRef`, so reassign the value instead.

The core-base test `recomputes the chain when the fallback is mutated in place` (added in #2564) now goes through `updateFallbackLocale()`, since core-base alone no longer detects the mutation.

### Commits

1. `fix(core-base)`: identity-keyed cache, and invalidation of the context and its root context
2. `fix(vue-i18n-core)`: keep one `fallbackLocale` reference in the core context, so `t()` and `te()` share a cache entry, and reach the root while invalidating from the setter
3. `fix(vue-i18n-core)`: deep watcher for in-place mutation

### Performance

`fallbackWithLocaleChain` cache-hit path, production bundle (`core-base.esm-browser.prod.js`), 1,000,000 calls, Node 24:

| `fallbackLocale` | before #2564 | master | this PR |
| --- | ---: | ---: | ---: |
| `'en'` | 4 ms | 91 ms | 6 ms |
| `['en', 'ja']` | 3 ms | 120 ms | 8 ms |
| map, 5 locales | 3 ms | 290 ms | 8 ms |
| map, 30 locales | 3 ms | 1,740 ms | 8 ms |
| two fallbacks in alternation (root / local) | — | 293 ms | 10 ms |

"before #2564" returns a stale chain when the fallback changes (#2562), so it is only a reference point.

Memory, 50,000 distinct object fallbacks, references dropped, after GC: master retains 14.7 MB (50,000 entries), this PR 0.6 MB (0 strong entries).

Size of `core-base.esm-browser.prod.js`: +56 bytes gzip.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed stale locale fallback chains after fallback locales are reassigned or modified in place.
  * Improved fallback behavior for local composers and inherited root configurations.
  * Ensured reactive fallback-locale changes are detected promptly, including deep object mutations.
  * Prevented cached fallback data from accumulating unnecessarily when multiple fallback configurations are used.
  * Preserved separate fallback chains when switching between different fallback configurations.


kazupon:
Thanks! This is a real bug, but it isn't introduced by this PR: `set inheritLocale`
has never written the inherited values back to the core context, on master and v11 alike.
It's tracked in #2621 and will be fixed in a separate PR. Syncing `fallbackLocale` alone
isn't enough, since `_context.locale` is stale as well.

PR #2641, opened 2026-09-21
fix(core-base): map locale unions without the UnionToTuple round-trip

### LINKED ISSUE #2640: TS2589 from LocaleMessages: the UnionToTuple round-trip is O(n²) and redundant (fails at ≥48 locales)
## Summary

`LocaleMessages`, `DateTimeFormats` and `NumberFormats` convert the `Locales` union to a
tuple and immediately index it back to a union. The tuple is never used as a tuple — every
consumer only reads `T[number]`. The round-trip costs O(n²) type instantiations and trips
`TS2589: Type instantiation is excessively deep and possibly infinite` once the locale union
reaches **48 members**.

Apps with a large locale list therefore cannot type-check at all. Replacing the round-trip
with a direct mapped type removes the error and cuts instantiations from **44,500 to 7**.

### Expected behavior

Apps with a large locale list do not encounter an error.

## Minimal reproduction (no Vue, no Nuxt, no components)

```ts
import type { UnionToTuple } from '@intlify/core-base'

type L49 =
  | 'a1'  | 'a2'  | 'a3'  | 'a4'  | 'a5'  | 'a6'  | 'a7'  | 'a8'
  | 'a9'  | 'a10' | 'a11' | 'a12' | 'a13' | 'a14' | 'a15' | 'a16'
  | 'a17' | 'a18' | 'a19' | 'a20' | 'a21' | 'a22' | 'a23' | 'a24'
  | 'a25' | 'a26' | 'a27' | 'a28' | 'a29' | 'a30' | 'a31' | 'a32'
  | 'a33' | 'a34' | 'a35' | 'a36' | 'a37' | 'a38' | 'a39' | 'a40'
  | 'a41' | 'a42' | 'a43' | 'a44' | 'a45' | 'a46' | 'a47' | 'a48'
  | 'a49'

export type T = UnionToTuple<L49>
//                ^ error TS2589: Type instantiation is excessively deep and possibly infinite.
```

In isolation, 48 members compile (21,748 instantiations) and 49 do not. Through `useI18n()`
the surrounding inference adds its own instantiations, so the cliff arrives one locale
earlier, at 48.

### As users actually hit it

```ts
import { useI18n } from 'vue-i18n'

declare module '@intlify/core-base' {
  interface GeneratedTypeConfig {
    locale: 'af' | 'am' | /* … 75 locales … */ | 'zu'
    legacy: false
  }
}

export function useStrings() {
  const { t } = useI18n() // error TS2589
  return t('some_key')
}
```

`@nuxtjs/i18n` generates exactly this `GeneratedTypeConfig` augmentation from the app's
configured locales, so every Nuxt app with ≥48 locales hits it. Reported there as
[nuxt-modules/i18n#3465](https://github.com/nuxt-modules/i18n/issues/3465) (open since
March 2025), but the defective type is here in `@intlify/core-base`.

## Root cause

```ts
// @intlify/core-base
type UnionToTuple<U, Last = LastInUnion<U>> =
  [U] extends [never] ? [] : [...UnionToTuple<Exclude<U, Last>>, Last]

type LocaleRecord<T extends any[], R> = { [K in T[number]]: R }

type LocaleMessages<Schema, Locales = Locale, Message = string> =
  LocaleRecord<UnionToTuple<Locales>, Schema>
```

`UnionToTuple` recurses once per union member, and each step runs `LastInUnion` (a
`UnionToIntersection` inference) over the remaining members — O(n²) instantiations, with no
depth guard.

The result is then only ever indexed with `[number]`, which throws the ordering away and
returns the original union. **All four uses of `UnionToTuple` in the package do this:**

| Location | Use |
| --- | --- |
| `LocaleMessages` | `LocaleRecord<UnionToTuple<Locales>, Schema>` → `{ [K in T[number]]: R }` |
| `DateTimeFormats` | `LocaleRecord<UnionToTuple<Locales>, Schema>` → same |
| `NumberFormats` | `LocaleRecord<UnionToTuple<Locales>, Schema>` → same |
| `FallbackLocales` | `PickupFallbackLocales<UnionToTuple<Locales>>` = `` T[number] \| `${T[number]}!` `` |

So the tuple is never load-bearing.

## Measurements

`tsc --extendedDiagnostics`, one file, `skipLibCheck: true`.

Scaling of `useI18n()` against locale-union size:

| Locales | Instantiations | TS2589 |
| --- | --- | --- |
| 5 | 5,297 | no |
| 20 | 9,812 | no |
| 35 | 18,152 | no |
| 45 | 25,837 | no |
| 47 | 27,578 | no |
| **48** | 27,126 | **yes** |
| 75 | 49,293 | yes |

Current type vs. proposed fix at 75 locales:

| Formulation | Instantiations | TS2589 |
| --- | --- | --- |
| `LocaleRecord<UnionToTuple<L>, Schema>` | 44,500 | yes |
| `{ [K in L & PropertyKey]: Schema }` | **7** | no |

## Proposed fix

Drop the tuple round-trip and map the union directly:

```ts
type LocaleRecord<Locales, R> = { [K in Locales & PropertyKey]: R }

type LocaleMessages<Schema, Locales = Locale, Message = string> =
  LocaleRecord<Locales, Schema>
```

and correspondingly:

```ts
type PickupFallbackLocales<T> = T | `${T & string}!`
```

This produces the same key set (`keyof` is identical), removes the error, and makes the cost
linear. `UnionToTuple` / `LastInUnion` can stay exported for compatibility, just unused on
these paths.

If preserving the tuple is desired for other reasons, an accumulator-based `UnionToTuple`
would at least make it O(n) rather than O(n²), but given every consumer immediately calls
`[number]`, removing it looks strictly better.

## Workaround for affected users

Passing explicit type arguments bypasses the defaulted `Locales`:

```ts
const { t } = useI18n<object, string>()
```

This does **not** cost message-key typing: with `DefineLocaleMessage` augmented, key
autocompletion still resolves (verified via the TypeScript language service — 831/831 keys
offered with and without the explicit type arguments). But it has to be applied at every
`useI18n()` call site, which is why a fix in the type itself is worth having.

### System Info

```shell
Versions: `vue-i18n@11.4.12`, `@intlify/core-base@11.4.12`, `typescript@5.9.3`.

  System:
    OS: Windows 11 10.0.26200
    CPU: (16) x64 AMD Ryzen 7 7840HS w/ Radeon 780M Graphics     
    Memory: 9.27 GB / 27.72 GB
  Binaries:
    Node: 22.17.1 - C:\Program Files\nodejs\node.EXE
    npm: 11.10.0 - C:\Program Files\nodejs\npm.CMD
    pnpm: 10.20.0 - \AppData\Roaming\npm\pnpm.CMD
  Browsers:
    Chrome: 153.0.8010.52
    Edge: Chromium (152.0.4191.62)
    Firefox: 156.0 - C:\Program Files\Mozilla Firefox\firefox.exe
  npmPackages:
    @intlify/core-base: ^11.4.12 => 11.4.12 
    @intlify/eslint-plugin-vue-i18n: 4.* => 4.5.1 
    @vitejs/plugin-vue-jsx: ^5.1.0 => 5.1.6 
    @vue/test-utils: ^2.4.11 => 2.5.1 
    vite: ^8.3.0 => 8.3.0 
    vitest: ^5.0.1 => 5.0.1 
    vue-tsc: ^3.3.1 => 3.3.11 
    vuetify: 3.8.* => 3.8.12 
    vuetify-nuxt-module: ^0.19.4 => 0.19.5
```

### Screenshot

_No response_

### Additional context

Experienced the problem in an app containing over 70 locales.

### Validations

lazerg:
### Description

Fixes #2640

`LocaleMessages`, `DateTimeFormats` and `NumberFormats` feed `UnionToTuple<Locales>` into `LocaleRecord`, and `FallbackLocales` feeds it into `PickupFallbackLocales`. Both only read `T[number]`, so the tuple is discarded as soon as it is built. `UnionToTuple` recurses once per locale and runs `LastInUnion` over the remaining members at every step, so it costs O(n²) instantiations and trips `TS2589` once the union reaches 48 locales. Apps with a locale list that long, which `@nuxtjs/i18n` generates from their config, cannot type-check at all.

Because only `T[number]` is read, passing the union wrapped in a one-element tuple produces the same key set without the recursion. `UnionToTuple`, `LastInUnion`, `LocaleRecord` and `PickupFallbackLocales` keep their signatures, so importers are unaffected, and the existing type tests already use that form (`PickupFallbackLocales<['en' | 'ja']>`).

At 75 locales, `tsc --extendedDiagnostics` over the four types goes from 46,942 instantiations with `TS2589` to 3,114 without.

The added test in `packages/core-base/test/types.test-d.ts` uses a 50-locale union and fails with `TS2589` on `master`.

## Summary by CodeRabbit

- **Improvements**
  - Improved TypeScript locale typing for message catalogs, date-time formats, number formats, and fallback locale definitions.
  - Locale unions are now preserved as single locale key types, providing more accurate type checking for projects supporting many locales.

- **Tests**
  - Added compile-time coverage confirming that large locale sets retain their expected key types across localization resources.


kazupon (changes requested):
Thank you for the PR! 🙏

The `[Locales]` approach is a nice call. It keeps `LocaleRecord` and `PickupFallbackLocales` unchanged, so we can back-port this to v11 as a patch.

One small request: `FallbackLocales` also changed here, but it doesn't have a large locale union test like the other three.

Could you add one? 🙏 Something like this:
```ts
describe('FallbackLocales', () => {
  test('large locale union', () => {
    type Locales = `${'p' | 'q' | 'r' | 's' | 't'}${1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10}`
    expectTypeOf<FallbackLocales<Locales>>().toEqualTypeOf<
      Locales | Locales[] | { [x in string]: (Locales | `${Locales}!`)[] } | false
    >()
  })
})
```

It passes with this PR and fails with TS2589 on master.

lazerg:
I added the `FallbackLocales` test in 8648fe4. It fails with TS2589 on master and passes with this PR.

